PR #100, opened 2015-06-30
Support both Python 2.7 and Python 3.x

satterly:
To run tests locally, for python 2.7 use:
```
python -m nose
```
And for python 3.4 use:
```
python3 -m nose
```


satterly:
Python 3.4 failing to build on Travis due to lack of support by librabbitmq. See celery/librabbitmq/issues/13

PR #102, opened 2015-07-10
Enable mongodb text search of all string content in alerts

satterly:
See [Text Indexes](http://docs.mongodb.org/manual/core/index-text/)
eg. http://localhost:8080/alerts?q={"$text":{"$search":"foo"}}


satterly:
Travis-CI still uses MongoDB version 2.4.12 which had text search included as a beta feature and was disabled by default.

Issue #111, opened 2015-09-04
the concept of 'maintenance mode'?


blysik:
Does Alerta have a mechanism for the concept of 'maintenance mode'?

ie, I'm going to do work on a server, which would cause alerts, so I want to temporarily mute alerts being generated from it.

satterly:
This has been on the backlog for a while... waiting for someone with a real use case to request it. Suggestions on how it might work in practice are welcome.

I'm looking at using a plugin to suppress the alert if it meets certain criteria -- but what criteria and how it is determined (and updated) is open to debate.

blysik:
I was thinking about this.  The first thing that came to mind is a list of regular expressions.  If an alert matches, note that it's in maintenance mode instead of doing the normal behavior.

alerta-api gets CRUD for maintenance mode expressions.  Python client and web-UI gets enhancements to deal with them.

satterly:
I'm reluctant to support a list of rules based on regular expressions because it then becomes a case of iterating through the list of rules, running the regex and seeing if any match. This could potentially be very bad for performance. Alternatively, you use an event stream processor but that is on another level of complexity.

The simple alternative, as far as I can tell, is to use equality match rules that work as a simple query. Alerta defines many different alert attributes that can be used to group alerts and it is these attributes that I have used to define rules eg. `environment`, `service`, `group`. However, `resource` and `event` attributes are supported for situations that require that level of granularity.

I have also supported the use of `tags` to define a blackout rule which should allow a lot of flexibility -- one or more tags can be required to match an alert for the suppression to apply. And tags can be added at source, using the `alerta` CLI, or using a plug-in.

I have only done the work on the API at present. Feedback on this would be welcome before I start work on adding support to the web UI and CLI tool.

blysik:
That makes sense.  

I think if my users could say something like, "mute all alerts where resource=hostname", that would work.  Or "service=some-service".

blysik:
The reason I initially thought of regexes, is if the users weren't certain what the 'resource' was going to look like.  ie, did they correctly use FQDN, or short name?  Or to make it easier to do something like mute hostname[1-30].company.com.

satterly:
I completely understand the requirement for something like `hostname[1-30].company.com` however this is very difficult to do efficiently. I would suggest that to achieve this you would tag those 30 hosts with whatever makes them common (eg. `frontend`) and use that tag to add/remove them from maintenance.

[...]

PR #112, opened 2015-09-07
Suppress alerts based on blackout rules to support maintenance mode

satterly:
An alert that is received during a blackout period is accepted by Alerta and a `202 Accepted` status code returned. However, it will not be added to the Alerta database.

Blackout rules can be any of:
  * an entire environment eg. `environment=Production`
  * a particular resource eg. `resource=host55`
  * an entire service eg. `service=Web`
  * every occurrence of a specific event eg. `event=DiskFull`
  * a group of events eg. `group=Syslog`
  * a specific event for a resource eg. `resource=host55` and `event=DiskFull`
  * all events that have a specific set of tags eg. `tags=[ blackout, london ]`

Note: an `environment` must be defined for a blackout rule.

Requested by #111

### DIFF
--- a/alerta/app/database.py
+++ b/alerta/app/database.py
@@ -930,14 +930,18 @@ class Mongo(object):
         elif service:
             data["priority"] = 3
             data["service"] = service
-        elif resource and event:
+        elif event and not resource:
             data["priority"] = 4
             data["event"] = event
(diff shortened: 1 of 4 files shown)


BCWilsonDotCom:
:+1:

PR #117, opened 2015-09-25
Update views.py

skob:
not needed to alive method -- see http://api.mongodb.org/python/current/api/pymongo/mongo_client.html


satterly:
Thanks.

satterly:
Just uploaded to [Pypi](https://pypi.python.org/pypi/alerta-server/4.5.1) as version 4.5.1.

PR #123, opened 2015-10-28
Update twitter auth callback endpoint to use POST not GET

satterly:
Satellizer.js auth library changed the OAuth1 callback to POST
to be consistent with OAuth2.

See https://github.com/sahat/satellizer/issues/504


satterly:
Satellizer.js updated https://github.com/alerta/angular-alerta-webui/pull/28

PR #150, opened 2016-02-15
Use alarm['AlarmName'] instead of alarm['Trigger']['MetricName'] for …

### LINKED ISSUE #149: Use alarm['AlarmName'] instead of alarm['Trigger']['MetricName'] for event in cloudwatch webhook
Motivation:
AWS MetricName (e.g. RequestCount, CPUUtilization, etc.) can be used to define many different CloudWatch Alarms. For example I want alarm when `RequestCount` is too small on ELB and when it is too high. These two alarm will have the same event name (`RequestCount`), the same resource (ELB),. 

I think using AlertName as an event is better. because it says what is the even.Plus you get to customize it in AWS Cloudwatch. 

PS: If you agree I can give PR.

tdi:
…event in cloudwatch webhook

Fixes #149

### DIFF
--- a/alerta/app/webhooks/views.py
+++ b/alerta/app/webhooks/views.py
@@ -58,7 +58,7 @@ def parse_notification(notification):
 
         return Alert(
             resource='%s:%s' % (alarm['Trigger']['Dimensions'][0]['name'], alarm['Trigger']['Dimensions'][0]['value']),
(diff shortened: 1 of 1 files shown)


satterly:
:+1:

satterly:
Alerta server version 4.7.2 now available for download.

PR #153, opened 2016-02-21
PUT on /user/<user> now updates all fields

### LINKED ISSUE #152: updating user (PUT /user/<user>) should update the whole user, not just password
Now only password is updated when you `PUT /user/<user>`. According to PUT semantics, I expected to update the whole user. There is no alternative way of updating users in alerta than that. 

I can provide a PR for this.

tdi:
I also made sure that backwards compatibility is maintained, so still you can update single field too, i.e. password. So we do not break angular and python UIs. Closes #152


alerta/app/views.py
```diff
@@ -724,20 +724,14 @@ def create_user():
...
+        user = db.update_user(user, request.json)
+        if user:
             return jsonify(status="ok")
         else:
             return jsonify(status="error", message="not found"), 404
-
-    else:
-        return jsonify(status="error", message="Must supply new password for user")
+    except Exception as e:
+        return jsonify(status="error", message=str(e)), 500
```
satterly: The exception should wrap the call to the database. Like so...

```
def update_user(user):

    try:
        user = db.update_user(user, request.json)
    except Exception as e:
        return jsonify(status="error", message=str(e)), 500

    if user:
        return jsonify(status="ok")
    else:
        return jsonify(status="error", message="not found"), 404
```

alerta/app/database.py
```diff
@@ -1108,6 +1108,40 @@ def is_user_valid(self, id=None, name=None, login=None):
         if login:
             return bool(self._db.users.find_one({"login": login}))
 
+
+    def update_user(self, login, fields=None):
```
satterly: This method needs to update the user by `id` not `login`. I suggest the method signature should look like this...

```
def update_user(self, id, name, login, password=None, provider="", text="", email_verified=False):
```

alerta/app/views.py
tdi: Thanks.

alerta/app/database.py
tdi: This change will break compatibility with `python-alerta`, which assumes that the functions. I am generally ok with this, since the server should enforce API changes. Are you ok with it?

tdi:
I have update the signature, changed views accordingly.

alerta/app/views.py
```diff
@@ -723,21 +723,35 @@ def create_user():
 @admin_required
 @jsonp
 def update_user(user):
-
+   
+    name, login, password, provider, text, email_verified = None, None, None, None, None, None
+
```
satterly: This can be replaced with something like ...
```
provider = request.json.get("provider", None)
```
... similar to how it's done in the `create_blackout()` method.

satterly:
LGTM other than the one comment above. Thanks.

tdi:
I fixed that. Thanks. Still learning the codebase.

satterly:
Good work. Thanks.

satterly:
See corresponding `alerta` CLI pull request https://github.com/alerta/python-alerta/pull/27

PR #154, opened 2016-02-21
Pattern match on Authorization header to get api-key/token

satterly:
Fix problem reported by @joaocavichiolli where if an authorization header is passed without a key but with the prefix "Key" or "Bearer" then it thinks the API key is "Key" or "Bearer".

```
2016-02-21 08:18:26,120 - requests.packages.urllib3.connectionpool - DEBUG - "POST /alert HTTP/1.1" 401 65
2016-02-21 08:18:26,121 - alerta.shell - ERROR - API key 'Key' is invalid
2016-02-21 08:18:26,121 - alerta.shell - WARNING - Exiting alerta client.
```

### DIFF
--- a/alerta/app/auth.py
+++ b/alerta/app/auth.py
@@ -3,6 +3,7 @@ import jwt
 import json
 import requests
 import bcrypt
+import re
 
 from datetime import datetime, timedelta
 from functools import wraps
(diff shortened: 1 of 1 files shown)


joaocavichiolli:
Hi Nick thanks for report.
Just substitute file auth.py in alert to fix ?

satterly:
If you wanted to patch it manually, then yes you should just replace `auth.py`.

PR #166, opened 2016-03-03
Add call to `apt-get upgrade`

adamnfish:
It looks like this might be missing a call to `apt-get upgrade`?


satterly:
Thanks :+1:

Issue #176, opened 2016-04-06
Allow closing alerts on blackout.


tdi:
An example: 
- alert X with severity critical comes,
- a blackout is set cacthing alert X,
- alert X comes with OK (but is neglected).
- blackout is off,
- alert X comes with critical, and becomes a duplicate. 

Maybe there should be a blackout (configurable via `BLACKOUT_ACCEPT=['ok', 'normal']`) that pass alerts of said severities. 

Just an idea. (We can provide a PR).

satterly:
Yes, this is a good suggestion. A PR would be very welcome. Thanks!

PR #177, opened 2016-04-07
Allow closing alterts while blackout with certain severities. Closes …

tdi:
…#176

Severities are configured via BLACKOUT_ACCEPT configuration option,
e.g. BLACKOUT_ACCEPT=['ok', 'normal']

### DIFF
--- a/alerta/app/database.py
+++ b/alerta/app/database.py
@@ -860,6 +860,9 @@ class Mongo(object):
         return blackouts
 
     def is_blackout_period(self, alert):
+        
+        if alert.severity in app.config.get('BLACKOUT_ACCEPT', []):
+            return False
 
         now = datetime.datetime.utcnow()
 


alerta/app/database.py
```diff
@@ -860,6 +860,11 @@ def get_blackouts(self, query=None):
         return blackouts
 
     def is_blackout_period(self, alert):
+        
+        blackout_accept = app.config.get('BLACKOUT_ACCEPT', None)
+        if blackout_accept:
+            if alert.severity in blackout_accept:
+                return False
```
satterly: I would simplify this as...
```
if alert.severity in app.config.get('BLACKOUT_ACCEPT', []):
    return False
```

tdi:
I have changed the code as you suggested.

satterly:
I assume you have tested this and it works as expected?

satterly:
:+1: Thanks

PR #184, opened 2016-04-22
Read settings from flask not directly from settings.py

satterly:
This was preventing `ALLOWED_ORIGINS` and `ALLOWED_ENVIRONMENTS` configuration settings from working correctly.


satterly:
Confirmed to work by @menardorama

PR #248, opened 2016-09-06
Fix versions for flask and flask_cors

satterly:
Should fix alerta/angular-alerta-webui#62


satterly:
Looks like "allows headers" list is now case sensitive so have raised an issue with Flask CORS corydolphin/flask-cors#187

PR #253, opened 2016-09-07
Set default switch status from config settings

satterly:
Can now set default values switch settings in the server config ...

```python
AUTO_REFRESH_ALLOW = 'OFF'
SENDER_API_ALLOW = 'ON'
```

### DIFF
--- a/alerta/app/management/views.py
+++ b/alerta/app/management/views.py
@@ -17,8 +17,8 @@ LOG = logging.getLogger(__name__)
 
 
 switches = [
-    Switch('auto-refresh-allow', 'Allow consoles to auto-refresh alerts', SwitchState.ON),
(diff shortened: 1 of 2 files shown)


PMDubuc:
Is there a way to modify the refresh interval and set it to some other value than 5 seconds?

Anderen2:
@PMDubuc Something like this: https://github.com/alerta/angular-alerta-webui/blob/master/app/config.json.example#L38 ?

satterly:
For version 6.0 it has moved to become a server setting

```
AUTO_REFRESH_INTERVAL = 5000  # ms
```
https://github.com/alerta/alerta/blob/3d9725355988327990420924b9dbc7b20e7c5b5d/alerta/settings.py#L148

PR #266, opened 2016-10-12
Treat unknown severity as normal for alert processing

satterly:
This is for cases where a normal alert is received for a resource and no notifications are wanted so the difference between this alert and a normal following a high severity alert is that the `trendIndication` will be `noChange` compare with `lessSevere`.

### DIFF
--- a/tests/test_severity.py
+++ b/tests/test_severity.py
@@ -1,0 +1,240 @@
+
+import unittest
+
+try:
+    import simplejson as json
+except ImportError:
+    import json
+
+from uuid import uuid4
+from alerta.app import app
+
+
+class SeverityTestCase(unittest.TestCase):
+
+    def setUp(self):
+
+        app.config['TESTING'] = True
+        app.config['AUTH_REQUIRED'] = False
+        self.app = app.test_client()
+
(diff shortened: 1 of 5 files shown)


tests/test_severity.py
```diff
@@ -0,0 +1,240 @@
...
+        # prevSev=unknown, sev=ok, status=closed, trend=noChange
+        response = self.app.post('/alert', data=json.dumps(self.ok_alert), headers=self.headers)
+        self.assertEqual(response.status_code, 201)
+        data = json.loads(response.data.decode('utf-8'))
+        self.assertEqual(data['alert']['duplicateCount'], 0)
+        self.assertEqual(data['alert']['repeat'], False)
+        self.assertEqual(data['alert']['previousSeverity'], 'unknown')
+        self.assertEqual(data['alert']['severity'], 'ok')
+        self.assertEqual(data['alert']['status'], 'closed')
+        self.assertEqual(data['alert']['trendIndication'], 'noChange')
```
satterly: This is the test where the `trendIndication` is shown to be `noChange` for `ok` alerts received when previous severity is `unknown`.

PR #269, opened 2016-10-21
add bind host option to alerta daemon

ilcic:
Add bind host option to alerta daemon instead of always binding 0.0.0.0

### DIFF
--- a/alerta/app/shell.py
+++ b/alerta/app/shell.py
@@ -21,6 +21,13 @@ def main():
         help='Listen port (default: 8080)'
     )
     parser.add_argument(
+        '-H',
+        '--host',
+        type=str,
+        default='0.0.0.0',
+        help='Bind host (default: 0.0.0.0)'
+    )
+    parser.add_argument(
         '--debug',
         action='store_true',
         default=False,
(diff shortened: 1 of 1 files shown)


alerta/app/shell.py
```diff
@@ -30,4 +37,4 @@ def main():
 
     LOG.info('Starting alerta version %s ...', __version__)
     LOG.info('Using MongoDB version %s ...', db.get_version())
-    app.run(host='0.0.0.0', port=args.port, debug=args.debug, threaded=True)
+    app.run(host='args.host', port=args.port, debug=args.debug, threaded=True)
```
satterly: Are those single quotes meant to be there?

ilcic: sorry that's an typo

satterly:
Thanks. Happy to merge this. Curious as to what your particular use case is though.

ilcic:
We don't want to bind the ip on virtual interface create by HAProxy.

satterly:
Makes sense. Alerta server version 4.8.8 is now available on Pypi with this feature. Thanks.

PR #271, opened 2016-10-21
record remote address of an alert

ilcic:
Since in our scenario we have multiple prometheus-alertmanagers, we want to record the remote ip of them.

### DIFF
--- a/alerta/app/alert.py
+++ b/alerta/app/alert.py
@@ -62,6 +62,7 @@ class Alert(object):
         self.timeout = kwargs.get('timeout', DEFAULT_TIMEOUT)
         self.raw_data = kwargs.get('raw_data', kwargs.get('rawData', None)) or ""
         self.customer = kwargs.get('customer', None)
+        self.remote = kwargs.get('remote', None)
 
     def get_id(self, short=False):
 
@@ -99,7 +100,8 @@ class Alert(object):
             'createTime': self.get_date('create_time', 'iso'),
             'timeout': self.timeout,
             'rawData': self.raw_data,
-            'customer': self.customer
+            'customer': self.customer,
+            'remote': self.remote
         }
 
     def get_date(self, attr, fmt='iso', timezone='Europe/London'):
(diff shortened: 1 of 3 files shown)


satterly:
This is already supported ... https://github.com/guardian/alerta/blob/master/alerta/app/views.py#L199

The remote IP address appears as an attribute in the `attributes` field ...
```
{
  "alerts": [
    {
      "attributes": {
        "ip": "208.36.133.238"
      },
```

satterly:
If you want to keep track of which prometheus alert manager sent an alert to Alerta then that is what the `origin` attribute is for ... https://github.com/guardian/alerta/blob/master/alerta/app/webhooks/views.py#L337

ilcic:
Does this key also exist in POST /webhooks/prometheus?

There is only `moreInfo` in my `attributes` field of an alert.

{
  "attributes": {
    "moreInfo": "<a href=\"http://control2.cloud.com:9090/graph#%5B%7B%22expr%22%3A%22up%20%3D%3D%200%22%2C%22tab%22%3A0%7D%5D\" target=\"_blank\">Prometheus Graph</a>"
  },

ilcic:
Also we want to treat alerts from different remote ip address as separate alerts since we have multiple openstack server located in different contries with the same alert rule monitored by prometheus.

satterly:
The remote IP address should be included in an alert submitted via a webhook. So that needs to be fixed.

To enable you to distinguish between different prometheus servers in each country I would either use the `customer` attribute (eg. `customer=italy`) or overload the `environment` attribute (eg. `environment=italy:Production`). In both cases the individual alerts will be correctly name spaced and considered as separate alerts for the same prometheus rule.

The `customer` would be defined by using API keys assigned to that customer. To set an environment for each country use the global config in the prometheus alertmanager, for example:

```
global:
  external_labels:
    environment: "italy:Production"
    service: "Prometheus"
    region: "eu-west-1"
```

ilcic:
Awesome! Thanks :)

Issue #203, opened 2016-05-24
Is there some way to integrate alerta with grafana?


rydeng:
is there an easy way to integrate alerta with grafna？ It is stuff like plugin used by grafana

satterly:
There will be ... https://github.com/grafana/grafana/issues/2209

rydeng:
thanks very much,satterly.

satterly:
I'm happy to  leave this issue open until the grafana work is complete as a reminder.

colomboipl:
Hello Grafana v4.0 is Out!
I want to test the Alert feature. Configure a Grafana Notification Webhook to send the alerta web ui.
Can you help me .

msupino:
adding this to views.py in app/wesocket

@app.route('/webhooks/grafana', methods=['OPTIONS', 'POST'])
@cross_origin()
@auth_required
def grafana():

    hook_started = webhook_timer.start_timer()
    alerts = parse_grafana(request.data)

    LOG.warning(alerts)

    for incomingAlert in alerts:

        LOG.warning(incomingAlert)
        if g.get('customer', None):
            incomingAlert.customer = g.get('customer')

        add_remote_ip(request, incomingAlert)

        try:
            alert = process_alert(incomingAlert)
        except RejectException as e:
            webhook_timer.stop_timer(hook_started)
            return jsonify(status="error", message=str(e)), 403
        except Exception as e:
            webhook_timer.stop_timer(hook_started)
            return jsonify(status="error", message=str(e)), 500

        webhook_timer.stop_timer(hook_started)

    if alert:
        body = alert.get_body()
        body['href'] = absolute_url('/alert/' + alert.id)
        return jsonify(status="ok", id=alert.id, alert=body), 201, {'Location': body['href']}
    else:
        return jsonify(status="error", message="insert or update of grafana check failed"), 500

def parse_grafana(check):
    check = json.loads(check)
    evalMatches = check['evalMatches']
    timeout = 60

    alerts = []

    for item in check['evalMatches']:
        LOG.warning(item['metric'])
        alerts.append(Alert(
            status='open',
            resource=item['metric'],
            event=check['state'],
            environment='Production',
            severity='critical',
            service=[check['ruleName']],
            text="Value:" + str(item['value']),
            origin='Grafana',
            event_type='Grafana_event',
            raw_data=item['value'],
            timeout=timeout,
            value='Error',
            group='group'
        ))

    return alerts

and adding this to grafana should work
[...]

PR #300, opened 2016-12-02
added webhook for grafana 4.0 alerting

msupino:
See #203


msupino:
added.. passes py2.7 tests, fails on 3.x, ill need to setup an env for testing on 3.x

satterly:
I'm not against integrating Alerta with Grafana however I'm not sure it's ready for prime-time yet.
There seems to be very limited amount of information passed in the webhook and this makes generating meaningful alerts extremely difficult.

For example, is it possible to determine what resource/host/service the alert relates to?

Some other questions/issues raised by others would go a long way to improving things...
  * Alerting: Rule Category / Tags? grafana/grafana#6552
  * Include opentsdb tag name values in alert body grafana/grafana#6808
  * Add message in the webhook request grafana/grafana#6725

Thoughts?

colomboipl:
Any news ?

satterly:
Hi @colomboipl I've not heard back from @msupino yet. Do you have an opinion?

msupino:
This might not be the best implementation for grafana, but it's a start,
and can be enhanced later, grafana is not host centric, a graph can be made
from different sources.. let's see if it can be used for real grafana
implementation, mine isn't really in use, but others might have better
reference

satterly:
Sounds reasonable. Merging this, then and will refactor (slightly) and fix tests in following PR.

PR #304, opened 2016-12-05
Fix broken PagerDuty webhook

excessivedemon:
PagerDuty webhook is currently broken - this is because the current webhook is unable to retrieve the correct alert tied to the incident.


satterly:
Thanks.

satterly:
This is available now for download from Pypi as Alerta server version 4.8.16. See also PR #305 so any tests you could do to confirm PagerDuty webhooks now work as expected would be appreciated. Thanks again.

excessivedemon:
Confirming PagerDuty webhooks are now working as expected with the latest release. Thanks man!

PR #318, opened 2016-12-27
Added a Riemann webhook

jamtur01:
Adds a webhook for Riemann. Makes no assumptions about the format of
your Riemann events and consumes standard events. If events are
decorated with additional metadata: tags, environment, group, etc then
these will be used.

```python
return Alert(
        resource='%s-%s' % (alert['host'], alert['service']),
        event=alert['service'],
        environment=alert.get('environment', 'Production'),
        severity=alert.get('state', 'unknown'),
        status=alert.get('state', 'unknown'),
        service=[alert['service']],
        group=alert.get('group', 'Performance'),
        text=alert.get('description', None),
        value=alert.get('metric', None),
        tags=alert.get('tags', None),
        origin='Riemann',
        raw_data=alert
    )
```

Added a test too.

### DIFF
--- a/alerta/app/webhooks/views.py
+++ b/alerta/app/webhooks/views.py
@@ -745,7 +745,6 @@ def parse_riemann(alert):
         event=alert['service'],
         environment=alert.get('environment', 'Production'),
         severity=alert.get('state', 'unknown'),
-        status=alert.get('state', 'unknown'),
         service=[alert['service']],
         group=alert.get('group', 'Performance'),
         text=alert.get('description', None),
--- a/alerta/app/webhooks/views.py
+++ b/alerta/app/webhooks/views.py
@@ -736,3 +736,57 @@ def telegram():
         return jsonify(status="ok")
     else:
         return jsonify(status="error", message="no callback_query in Telegram message"), 400
+
+
+def parse_riemann(alert):
+
+    return Alert(
+        resource='%s-%s' % (alert['host'], alert['service']),
+        event=alert['service'],
+        environment=alert.get('environment', 'Production'),
+        severity=alert.get('state', 'unknown'),
+        status=alert.get('state', 'unknown'),
+        service=[alert['service']],
+        group=alert.get('group', 'Performance'),
+        text=alert.get('description', None),
+        value=alert.get('metric', None),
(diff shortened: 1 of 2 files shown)


jamtur01:
Please excuse anything egregious. This is the first Python I've written in ten years.

satterly:
Thanks for this. Your python is fine :)

One suggestion I would give is to *not* set the `status` to the value of `state` (as it is already used for `severity` anyway) and let the Alerta server determine the status based on the assigned severity. The status will be set by the Alerta server to either `closed`, if the passed severity is `ok`, and a status of `open` for any other severity.

Just an FYI... there is already a Riemann integration that can be included on the Riemann side but your approach of adding a webhook is just as valid and perhaps easier for those to get going more quickly. See https://github.com/alerta/riemann-alerta/blob/master/alerta.clj

jamtur01:
@satterly Not quite sure I am following re `status` and `state`? So....

1. Set the `severity` to `state`.

2. If the `severity` set is is `ok` then set state to `closed`, else set `status` to open?

Something like:


```python
    severity = alert.get('state', 'unknown')
    if severity == 'ok':
      status = 'closed'
    else:
      status = 'open'
    elif status == 'closed':
```

And yes. I saw the integration and hacked on it a bit (added support to pass an API token that I'll try to PR sometime this week). It feels more natural to maintain the event API locally to Alerta rather than downstream in Riemann hence decided to add this.

alerta/app/webhooks/views.py
```diff
@@ -736,3 +736,57 @@ def telegram():
...
+
+
+def parse_riemann(alert):
+
+    return Alert(
+        resource='%s-%s' % (alert['host'], alert['service']),
+        event=alert['service'],
+        environment=alert.get('environment', 'Production'),
+        severity=alert.get('state', 'unknown'),
+        status=alert.get('state', 'unknown'),
```
satterly: Delete this line.

jamtur01:
Ah the server setting it part I missed. Will update shortly. Thanks. :)

jamtur01:
Updated. Thanks!

satterly:
Nice one. Thanks.

PR #332, opened 2017-01-09
print<space> to print() (watchdog.py)

skob:


satterly:
I assume this is to make it work with Python 3.

satterly:
👍

skob:
Thank you :-)

PR #345, opened 2017-03-13
Allow configuration of SSL for MongoDB.

christian-marie:
Client side SSL certs can now be configured by setting a combination of:

MONGO_SSL_CA='/path/to/ca.pem'
MONGO_SSL_CERT='/path/to/cert.pem'
MONGO_URI='mongodb://CN=user01@node01,node02/monitoring?replicaSet=monitoring&authMechanism=MONGODB-X509'

Tested with mongodb 3.4.2

### DIFF
--- a/alerta/app/database/mongo.py
+++ b/alerta/app/database/mongo.py
@@ -5,6 +5,7 @@ import base64
 import hmac
 import hashlib
 import bcrypt
+import ssl
 
 from uuid import uuid4
 from six import string_types
@@ -49,8 +50,25 @@ class Database(object):
 
         mongo_uri = mongo_uri or app.config['MONGO_URI']  # use app config if no env var overrides
 
+        # Validate SSL options
+        to_verify = ['MONGO_SSL_CERT', 'MONGO_SSL_CA']
+        ssl_opts = { k: os.environ.get(k, False) or app.config.get(k, False) for k in to_verify }
+        missing_ssl_opts = { k:v for k, v in ssl_opts.items() if not v }
+
+        # All missing or none missing
+        if len(missing_ssl_opts) < len(ssl_opts) and len(missing_ssl_opts) != 0:
+            LOG.error('Missing option(s): {} '.format(", ".join(missing_ssl_opts)))
+            sys.exit(1)
+
+        ssl_args = {}
+        if len(missing_ssl_opts) == 0:
+            ssl_args = { 'ssl': True,
+                         'ssl_certfile': ssl_opts['MONGO_SSL_CERT'],
+                         'ssl_ca_certs': ssl_opts['MONGO_SSL_CA'],
+                         'ssl_cert_reqs': ssl.CERT_REQUIRED }
+
         try:
-            self.connection = MongoClient(mongo_uri, serverSelectionTimeoutMS=2000, connect=False)
+            self.connection = MongoClient(mongo_uri, serverSelectionTimeoutMS=2000, connect=False, **ssl_args)
         except Exception as e:
             LOG.error('MongoDB Client: %s : %s', mongo_uri, e)
             sys.exit(1)


satterly:
Thanks. It's a good idea. Tests for python 3.x are failing though. I'll take a look when I get a chance.

alerta/app/database/mongo.py
```diff
@@ -49,8 +50,30 @@ def connect(self):
...
 
+        # Validate SSL options
+        def try_find(k):
+            try:
+                return (os.environ.get(k) or app.config[k])
+            except KeyError: pass
+
+        to_verify = ['MONGO_SSL_CERT', 'MONGO_SSL_CA']
+        ssl_opts = { k: try_find(k) for k in to_verify }
+        missing_ssl_opts = { k:v for k, v in ssl_opts.iteritems() if v is None }
```
satterly: This seems overly complicated for what it does (and it doesn't work for python 3.x). How about something like this?
```
        ssl_args = {}
        ssl_cert = os.environ.get('MONGO_SSL_CERT', False) or app.config.get('MONGO_SSL_CERT', False)
        ssl_ca = os.environ.get('MONGO_SSL_CA', False) or app.config.get('MONGO_SSL_CA', False)

        if (ssl_cert and not ssl_ca) or (ssl_ca and not ssl_cert):
            LOG.error('Mongo SSL setup failed - must define MONGO_SSL_CERT and MONGO_SSL_CA')
            sys.exit(1)

        if ssl_cert and ssl_ca:
            ssl_args = {
                'ssl': True,
                'ssl_certfile': ssl_cert,
                'ssl_ca_certs': ssl_ca,
                'ssl_cert_reqs': ssl.CERT_REQUIRED
            }
```

christian-marie: The python 3x issue I can address, I suspect it's just s/iteritems/items/. 

As for this being overly complicated, that's your call. I didn't know that python had .get(x, False) for defaulting, so I can remove the helper function which certainly reduces perceived complexity.

1. 

If you want to go with your explicit version, I'd observe: (P ∧ ¬Q) ∨ (Q ∧ ¬P) == P ⊻ Q. So, it's just an xor which I find less verbose. The reason I didn't do this boolean statement was that the user doesn't know which variable is missing, and the solution is now specific to two arguments. As a general rule, if I have to deal with multiple elements I handle them as n elements and print errors nicely.

This code can later be pulled out into a helper to handle the general case of validating groups of n config items, and we aren't manually querying the underlying sources of configuration in multiple ways with different strings, all of which could be subtly wrong and will require manual replacement in multiple places if refactored.

Up to you. Shall I rewrite this with the explicit boolean check?

christian-marie:
I suspect my earlier comment was lost, I'll repeat here:

```
The python 3x issue I can address, I suspect it's just s/iteritems/items/.

As for this being overly complicated, that's your call. I didn't know that python had .get(x, False) for defaulting, so I can remove the helper function which certainly reduces perceived complexity.

If you want to go with your explicit version, I'd observe: (P ∧ ¬Q) ∨ (Q ∧ ¬P) == P ⊻ Q. So, it's just an xor which I find less verbose. The reason I didn't do this boolean statement was that the user doesn't know which variable is missing, and the solution is now specific to two arguments. As a general rule, if I have to deal with multiple elements I handle them as n elements and print errors nicely.

This code can later be pulled out into a helper to handle the general case of validating groups of n config items, and we aren't manually querying the underlying sources of configuration in multiple ways with different strings, all of which could be subtly wrong and will require manual replacement in multiple places if refactored.
```

There is a boolean version in #346 if you prefer.

Closes #346.

christian-marie:
We'll need documentation updated somewhere, yes?

christian-marie:
Also, it just occurred to me that it makes sense to require both CA and cert in my case, but perhaps not in general. Would you prefer if we allowed either to be specified in isolation, and then simply switch ssl: True if either are specified? Maybe we want to allow ssl: True by itself too! I can't  guarantee all permutations make sense, is all. CA=ca.pem SSL=false probably doesn't make sense for example.

Issue #347, opened 2017-03-16
Feature: support for Keycloak


m4ce:
It would be handy if Alerta had support for a generic OpenID/OAuth authenticator such as Keycloak. This would also solve the problem of connecting Alerta to LDAP since Keycloak can take care of that.

m4ce:
I have actually written it myself and now testing it. Once done, I will open a PR.

satterly:
Available now as Alerta server version 4.9.0.

PR #348, opened 2017-03-16
Adding support for Keycloak auth provider

m4ce:
Solves #347.

I am using roles (rather than groups) to manage the application permissions.


satterly:
Excellent work. Thanks a lot. 👍

PR #353, opened 2017-05-13
Setting default instance for Prometheus webhook

shredder12:
Instance is associated with every scrape target in Prometheus. But label isn't necessary for all alerts and may not even make sense for some. For instance, while alerting on a service level rule.

I've currently set it to 'NA'. If there is a better alternative, please suggest.

### DIFF
--- a/alerta/app/webhooks/views.py
+++ b/alerta/app/webhooks/views.py
@@ -340,7 +340,7 @@ def parse_prometheus(alert, external_url):
         annotations['moreInfo'] = '<a href="%s" target="_blank">Prometheus Graph</a>' % alert['generatorURL']
 
     return Alert(
-        resource=labels.pop('exported_instance', None) or labels.pop('instance'),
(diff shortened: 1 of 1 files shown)


satterly:
The `resource` field needs to be set to something useful otherwise you will get all alerts with the same `event` correlating / de-duplicating with each other (because the `resource` is the same `NA` value).

eg. `service=Email`, `event=Availability`, `resource=NA`
`service=DNS`, `event=Availability`, `resource=NA`

The two alerts above will be correlated and you will only see one. This is almost certainly what you don't want.

Can the label `instance` be set to the service name (or something uniquely relevant to that service) for service level alert rules?

eg. `service=Email`, `event=Availability`, `resource=EmailService` (instead of `resource=NA`)

satterly:
Closing. Please re-open if you want to discuss further.

shredder12:
Thank you @satterly. This makes sense.

PR #354, opened 2017-05-23
Adds support for event field

jamtur01:
Couple of folks gave me feedback that they set an event field in
Riemann when sending to Alerta and other destinations. The current
webhook discards this potential field and uses service instead. This
changes the behavior so that if the event field exists it is used
otherwise it defaults to service.


satterly:
Looks good. Thanks! 👍

satterly:
Available immediately as [Alerta server version 4.9.2](https://pypi.python.org/pypi/alerta-server/4.9.2)

PR #355, opened 2017-06-07
Use '*' to allow user role to see all customer

satterly:
Currently only admins can see all customers. This change will allow setting customer as "*" so that a user role may see all customers, but still not have admin rights.

In the example below the "noc@foo.com" user will be able to see all customer alerts but "user@foo.com" will only be able to see alerts for customer "Foo".

screen shot 2017-06-07 at 11 49 58


satterly:
Available as Alerta server version 4.9.3

PR #362, opened 2017-06-30
All users can manage alert blackouts

satterly:
>currently only admin users can create blackouts, would it be possible to make this available to regular users?

/cc @ramshankarjaiswal


ramshankarjaiswal:
I've just tested this, regular users still cannot see the blackout option in the alerta user interface.

satterly:
You're right. I thought I made the web UI change as well but don't see it. I'll take a look.

ramshankarjaiswal:
Thanks!

satterly:
Web UI updated to support user blackouts.

PR #364, opened 2017-07-05
SAML 2.0 Authentication

m-khvoinitsky:
This is a preliminary pull request containing SAML 2.0 authentication support. Please, review and comment if you have any concerns. Please, do not merge until complete WebUI implementation will be ready (just in case if something will be changed).

Additional changes:
* refactored absolute_url to produce correct results in some cases including calling it outside request context. Changed behaviour: the function called without arguments now always return url with trailing slash.
* (d281a68e539c7e2e9e82bba7b60cba82c468ae46) add "attrs" key to g object which is filled with attributes received from authn_response.get_identity().

alerta-doc PR alerta/alerta-docs#47

### DIFF
--- a/alerta/app/auth.py
+++ b/alerta/app/auth.py
@@ -11,11 +11,20 @@ except ImportError:
 
 from datetime import datetime, timedelta
 from functools import wraps
-from flask import g, request, render_template, jsonify
+from flask import g, request, render_template, jsonify, make_response
 from flask_cors import cross_origin
 from jwt import DecodeError, ExpiredSignature, InvalidAudience
 from base64 import urlsafe_b64decode
 from uuid import uuid4
+try:
+    import saml2
+    import saml2.entity
+    import saml2.metadata
+    import saml2.config
+    import saml2.client
+    import saml2.saml
+except ImportError:
+    pass  # saml2 authentication will not work
 
 import smtplib
 import socket
@@ -29,7 +38,7 @@ except ImportError:
     from urllib import urlencode
 
 from alerta.app import app, db
-from alerta.app.utils import absolute_url
+from alerta.app.utils import absolute_url, deepmerge
 
 BASIC_AUTH_REALM = "Alerta"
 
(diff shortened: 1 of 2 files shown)


satterly:
Can you supply an example of what might be contained in the `authn_response.get_identity()` that would be stored in the new `attrs` object?

m-khvoinitsky:
Sure, in my case it's:
```python
{'givenName': ['Mikhail'], 'surname': ['Khvoinitsky'], 'emailAddress': ['m_khvoinitsky@example.com'], 'group': ['alerta_rw']}
```
Exact returned claims are configured on the IdP side and can be altered by map from SAML2_CONFIG['attribute_map_dir'].
Main purpose of this is the ability to do some sort of ACL using pre_receive hooks.

satterly:
I'm not sure why you need the `attrs` object at all. Why do you need to persist this info in the database and provide it in the `g` request context?

And you don't seem to use the `group`. Why wouldn't you match against the SAML user `group`to  a list of `ALLOWED_SAML_GROUPS` like for the other auth types? eg. `ALLOWED_GITHUB_ORGS` and `ALLOWED_GITLAB_GROUPS`.

m-khvoinitsky:
> Why wouldn't you match against the SAML user groupto a list of ALLOWED_SAML_GROUPS like for the other auth types? eg. ALLOWED_GITHUB_ORGS and ALLOWED_GITLAB_GROUPS.

Well, the whole point of this was making it as abstract (i. e. not tied to my company's IdP) as possible. However I can change it to something like ALLOWED_SAML_GROUPS which should be more straightforward.

> I'm not sure why you need the attrs object at all. Why do you need to persist this info in the database and provide it in the g request context?

As I already wrote, main purpose of this is the ability to do some sort of ACL using pre_receive hooks in plugins. For this I need only 'groups' here but, again, I was trying to make it as abstract as possible. What about putting only 'groups' here if it available? However, in this case it will intersect with 'role'. Any thoughts?

satterly:
I'm all for making it as generic as possible and I can see now that I've looked again at your code that that was your intention. However, I think it would be good to be as consistent as possible with existing Alerta auth concepts while taking into account the specifics of SAML.

So, if `group` is not a standard SAML attribute then maybe you could make that configurable. ie. allow setting a `SAML_GROUP_ATTR` that can then be used to get an attribute to match against the list of `ALLOWED_SAML_GROUPS` to allow/deny access to Alerta.

In your case it would be `SAML_GROUP_ATTR=group` (which could be the default) and `ALLOWED_SAML_GROUPS=['alerta_rw']`.

Essentially I'm just suggesting breaking this up into two config variables ...
```
ALLOWED_SAML2_ATTRS = {
    'group': ['alerta_ro', 'alerta_rw']
}
```
... to ...
```
SAML_GROUP_ATTR='group'  # default
ALLOWED_SAML_GROUPS=['alerta_ro', 'alerta_rw']
```

m-khvoinitsky:
> So, if group is not a standard SAML attribute

Sort of. In my case this field is actually called `http://schemas.xmlsoap.org/claims/Group` and, for example, email field is called `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress`. To make it more human-friendly, pysaml2 uses so-called attribute_maps which can (and should be) configured. That's why if you have something like groups in your claims, you can always make it to be called `'groups'`. That's why `SAML2_GROUP_ATTR` parameter from your example is redundant. In the rest, I agree with you and will rewrite it to `ALLOWED_SAML2_GROUPS`.

So, what do you think about `role` vs. `groups` in `g` and JWT token? The problem is that current alerta roles is not flexible enough but adding competing groups would be inconsistent. I'm thinking about refactoring roles somehow to be more flexible but do not have solution yet.

satterly:
What are your requirements for roles that isn't currently met? A "read only" role?

m-khvoinitsky:
Yes. And it should be based on groups, not received from config file.

satterly:
I can't think of a simple solution to that right now that doesn't involve horrible hacks using Apache or something to deny PUT, POST, and DELETE methods based on some header. Let me think about it some more.

In the meantime I'd be happy to accept your SAML auth PR given the changes we discussed above and we can address the read-only user role separately. It has much wider implications than just SAML so it makes sense deal with it outside this PR.

m-khvoinitsky:
Agreed.
I'll create separate issue for discussing read-only role later.

satterly:
I've added scope-based roles to Alerta (see #367). This means it should be possible to define things like "read-only" roles (ie. set scope="read') or allow some users to delete alerts (add scope="admin:alerts"). Please rebase your PR with latest master to take advantage of this.

m-khvoinitsky:
I've updated the code (and documentation, see alerta/alerta-docs#47) according to your suggestion.

Web UI Pull Request: alerta/angular-alerta-webui#82

One more concern: I don't find it quite elegant how API transfers JWT token to WebUI. However, it is the best way I found to do this without WebUI restart. Do you have some thoughts on this?

satterly:
Good work. Thanks! There may be some conflicts when I merge in the "scopes" PR but I'll deal with them. 👍

PR #365, opened 2017-07-06
Fix Stackdriver webhook

nabilhassein:
This commit fixes two minor but breaking issues with the stackdriver webhook view:
1. The parse_stackdriver method was returning a tuple of a state (string) and an Alert,
where calling code was expecting it to return only an Alert.
2. The old code assumed that Stackdriver was setting the header
"Content-Type: application/json". Stackdriver should indeed set this, but
in practice it doesn't, so using request.get_json(force=True)
instead of request.json is an easy workaround.

### DIFF
--- a/alerta/app/webhooks/views.py
+++ b/alerta/app/webhooks/views.py
@@ -433,7 +433,7 @@ def parse_stackdriver(notification):
         status = None
         create_time = None
 
-    return state, Alert(
+    return Alert(
         resource=incident['resource_name'],
         event=incident['condition_name'],
         environment='Production',
@@ -461,7 +461,7 @@ def stackdriver():
 
     hook_started = webhook_timer.start_timer()
     try:
-        incomingAlert = parse_stackdriver(request.json)
+        incomingAlert = parse_stackdriver(request.get_json(force=True))
     except ValueError as e:
         webhook_timer.stop_timer(hook_started)
         return jsonify(status="error", message=str(e)), 400


satterly:
Good work. Thanks! It would be great to have some tests around this.

So, are you able to either add a test to `tests/test_webhooks.py` for stackdriver, or provide me with the raw json for a few requests (an open, acknowledged, closed) ?

nabilhassein:
Acknowledging the alert in stackdriver appears to have no effect on the alerta dashboard we have up, but I am able to pass along raw json for open and closed states!

Open:
```
{"incident": {"incident_id": "0.kenwzbdx55pj","resource_id": "","resource_name": "project-name Global labels { }","state": "open","started_at": 1499368214,"ended_at": null,"policy_name": "Alerta Whats Broken testing","condition_name": "Metric Threshold on Custom Metrics","url": "https://app.google.stackdriver.com/incidents/0.kenwzbdx55pj?project=project-name","summary": "custom/test.alerting for project-name Global labels { } is above the threshold of 0.5 with a value of 1."},"version": "1.1"}
```

And closed:
```
{"incident": {"incident_id": "0.kenwzbdx55pj","resource_id": "","resource_name": "project-name Global labels { }","state": "closed","started_at": 1499368214,"ended_at": 1499368836,"policy_name": "Alerta Whats Broken testing","condition_name": "Metric Threshold on Custom Metrics","url": "https://app.google.stackdriver.com/incidents/0.kenwzbdx55pj?project=project-name","summary": " custom/test.alerting for project-name Global labels { } returned to normal with a value of 0."},"version": "1.1"}
```

satterly:
Thanks. I've used the json to create some tests for Stackdriver. 👍

satterly:
This is available immediately as Alerta server version 4.9.6.

PR #367, opened 2017-07-08
Scope-based permissions model

satterly:
<table>
<tr><th>Name</th><th>Description</th></tr>
<tr><td><code>read</code></td><td>Grants read-only access to all scopes.</td></tr>
<tr><td><code>write</code></td><td>Grants read/write access to all scopes.</td></tr>
<tr><td><code>admin</code></td><td>Grants admin, read, write and delete access to all scopes.</td></tr>
<tr><td><code>read:alerts</code></td><td>Read-only access to alerts.</td></tr>
<tr><td><code>write:alerts</code></td><td>Grants read/write access to alerts.</td></tr>
<tr><td><code>admin:alerts</code></td><td>Grants read, write and delete access to alerts.</td></tr>
<tr><td><code>read:blackouts</code></td><td>Grants read-only access to blackouts.</td></tr>
<tr><td><code>write:blackouts</code></td><td>Grants read/write access to blackouts.</td></tr>
<tr><td><code>read:heartbeats</code></td><td>Read-only access to heartbeats.</td></tr>
<tr><td><code>write:heartbeats</code></td><td>Grants read/write access to heartbeats.</td></tr>
<tr><td><code>admin:heartbeats</code></td><td>Grants read, write and delete access to heartbeats.</td></tr>
<tr><td><code>admin:users</code></td><td>Fully manage users.</td></tr>
<tr><td><code>admin:customers</code></td><td>Fully manage customers.</td></tr>
<tr><td><code>read:keys</code></td><td>List and view API keys.</td></tr>
<tr><td><code>write:keys</code></td><td>Create, list and view API keys.</td></tr>
<tr><td><code>admin:keys</code></td><td>Fully manage API keys.</td></tr>
<tr><td><code>write:webhooks</code></td><td>Grants write access to webhooks.</td></tr>
<tr><td><code>read:oembed</code></td><td>Grants read-only to oembed endpoints.</td></tr>
<tr><td><code>read:management</code></td><td>Grants read-only access to management endpoints.</td></tr>
<tr><td><code>admin:management</code></td><td>Fully manage management endpoints.</td></tr>
<tr><td><code>read:userinfo</code></td><td>Grants read-only access to userinfo.</td></tr>
</table>

Note: `write` implicitly includes `read`, and `admin` implicitly includes `read` and `write`.

See alerta/angular-alerta-webui#81 and alerta/python-alerta-client#63

PR #368, opened 2017-07-13
Deprecate roles for scopes & permissions

satterly:
Following #367 this deprecates the use of "role" for all auth methods except basic auth.

### DIFF
--- a/alerta/app/database/mongo.py
+++ b/alerta/app/database/mongo.py
@@ -1213,7 +1213,6 @@ class Database(object):
     def get_users(self, query=None, password=False):
 
         users = list()
-
         for user in self.db.users.find(query):
             login = user.get('login', None) or user.get('email', None)  # for backwards compatibility
             u = {
@@ -1222,6 +1221,7 @@ class Database(object):
                 "login": login,
                 "createTime": user['createTime'],
                 "provider": user['provider'],
+                "role": 'admin' if login in app.config.get('ADMIN_USERS') else user.get('role', 'user'),
                 "text": user.get('text', ""),
                 "email_verified": user.get('email_verified', False)
             }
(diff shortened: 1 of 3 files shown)


alerta/app/database/mongo.py
```diff
@@ -1339,25 +1338,20 @@ def delete_user(self, id):
 
         return True if response.deleted_count == 1 else False
 
-    def get_role_by_match(self, matches):
+    def get_scopes_by_match(self, matches):
         if matches[0] in app.config['ADMIN_USERS']:
-            return 'admin'
+            return ['admin', 'read', 'write']
 
         if isinstance(matches, string_types):
```
m-khvoinitsky: This probably should be in the beginning of the function

alerta/app/database/mongo.py
```diff
@@ -1339,25 +1338,20 @@ def delete_user(self, id):
...
-        results = [find_role(m) for m in matches]
-        return next((r for r in results if r is not None), 'user')
+                return response['scopes']
 
-    def get_scopes(self, role):
-        role = self.db.roles.find_one({"role": role})
-        if role:
-            return role['scopes']
+        results = [find_scopes(m) for m in matches]
+        return next((r for r in results if r is not None), app.config['USER_DEFAULT_SCOPES'])
```
m-khvoinitsky: It took some time for me to understand this code :)
I'd suggest something like this:
```python
for q in matches:
    response = self.db.scopes.find_one({"match": q}, projection={"scopes": 1, "_id": 0})
    if response:
        return response['scopes']
return app.config['USER_DEFAULT_SCOPES']
```
Anyway, you return whether first result or default if there are no results.

m-khvoinitsky: By the way, if we have, say, matches = ['user', 'alerta_ro', 'alerta_rw'] — in this case the function will return RO scopes. It seems logical to merge scopes from all results instead. However, it will prevent you from revoking some scopes for certain user.

satterly: Yes, I had considered this and agree that merging scopes is a better approach. Will update.

PR #381, opened 2017-08-31
Allow specifying external link for apps like kibana

pbabics:
## Usecase
Alerts often correlate with other events which may not be logged in alerta, for this purpose it would be nice to allow specifying external link which leads to location containing more information, about event which produced specific alert and events that happened at this time


satterly:
Thanks for the PR but this is already possible using "custom attributes". To support your requirement you could define a custom attribute called "externalUrl" and the value would be the HTML link. There a quite a few examples of how this works in the [code to support webhooks](https://github.com/guardian/alerta/blob/master/alerta/app/webhooks/views.py#L448).

```
        attributes={
            'incidentId': incident['incident_id'],
            'resourceId': incident['resource_id'],
            'moreInfo': '<a href="%s" target="_blank">Stackdriver Console</a>' % incident['url']
        },
```

PR #384, opened 2017-09-06
alert can have zero(0) value

arindamchoudhury:
kwargs.get('value', None) or "n/a" sets value to 'n/a' if provided value is zero


satterly:
You are right that this isn't handled correctly. In the next release "value" will be cast to a string so this shouldn't be an issue shortly.

I'll merge this as is, though a more succinct fix would be ...
```
self.value = kwargs.get('value', "n/a")
```
Thanks for contributing back to the project.

PR #387, opened 2017-09-08
Add new webhook endpoint for slack integration

bcanyelles:
Add new endpoint for slack webhook.


satterly:
I'll assume this works as written though a test would have been nice ;)

I'm close to making Release 5 available and so will port across what you've done here to my development branch. I'll probably need to refactor a few things though hopefully not too much. I'll probably add a test to ensure I haven't broken what you've done.

Thanks for the PR.

PR #395, opened 2017-09-24
Release 5.0.0 RC1

satterly:
This branch is live on PyPi as `alerta-server 5.0.0rc1` https://pypi.python.org/pypi/alerta-server/5.0.0rc1


satterly:
👍

PR #405, opened 2017-10-18
Make housekeeping delete thresholds configurable params

### LINKED ISSUE #404: Allow for configurable housekeeping?
Moving from 4 to 5 the API endpoint for housekeeping is a nice change, but would it be possible to allow for some configuration (either in config file or via the URL)?

We have some alerts that fire fairly infrequently...in particular we have user laptop backups that send an OK on success.  This generally happens once a day, so with the default housekeeping they would get deleted fairly regularly.

The issue this causes is that the next day when the backup works and an OK is sent a Slack post is triggered (whereas if they are not deleted it would just be counted as a duplicate).  In v4 we just modified the suggested housekeeping script to expire closed alerts after 25 hours, but with housekeeping now internal to Alerta that's a little more cumbersome.

We could also modify our alert flow (we could easily just skip posting to slack on the success in this case), but this seems like something worth considering.

I believe the code at issue (for Postgres) is:

```
DELETE FROM alerts
    WHERE (status IN ('closed', 'expired') AND last_receive_time < (NOW() at time zone 'utc' - INTERVAL '2 hours'))
    OR (severity='informational' AND last_receive_time < (NOW() at time zone 'utc' - INTERVAL '12 hours'))
```

We don't need infinite flexibility, using variables for the `2` and the `12` would be plenty.

Thanks.

satterly:
**Example**
```
$ http ':8080/management/housekeeping?info=1&expired=1'
```

Fixes #404

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -991,14 +991,16 @@ class Backend(Database):
 
     #### HOUSEKEEPING
 
-    def housekeeping(self):
-        # delete 'closed' or 'expired' alerts older than 2hrs and 'informational' alerts older than 12hrs
-        two_hours_ago = datetime.utcnow() - timedelta(hours=2)
-        g.db.alerts.remove({"status": {'$in': ["closed", "expired"]}, "lastReceiveTime": {'$lt': two_hours_ago}})
+    def housekeeping(self, expired_threshold, info_threshold):
+        # delete 'closed' or 'expired' alerts older than "expired_threshold" hours
+        # and 'informational' alerts older than "info_threshold" hours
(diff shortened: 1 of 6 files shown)


thehilll:
Is there a way to install this version yet?  When I do `pip install alerts-server upgrade` I get:

```
Could not find a version that satisfies the requirement upgrade (from versions: )
No matching distribution found for upgrade
```

Thanks.

satterly:
I've just cut a new release for this - version 5.0.1 is available now.

And I think you were mis-typed the command:

```
$ pip install alerta-server --upgrade
```
or
```
$ pip install alerta-server==5.0.1
```
should work.

thehilll:
Thanks, got it.

PR #409, opened 2017-10-19
auth/google: fallback to email if profile['name'] not available

icy:
Sometimes, `profile['name']` is not available (at least, in our company settings). This patch uses email as a fallback value. We may have better fallbacks (e.g, by removing domain from email,...). This PR is an idea. Please let me know if I can help.

NB: The fallback code is from https://github.com/redlotus. I prefer Ruby way 🍡

### DIFF
--- a/alerta/auth/google.py
+++ b/alerta/auth/google.py
@@ -47,6 +47,6 @@ def google():
 
     customer = get_customer(id_token.email, groups=[domain])
 
-    token = create_token(id_token.subject, profile['name'], id_token.email, provider='google', customer=customer,
+    token = create_token(id_token.subject, 'name' in profile and profile['name'] or id_token.email, id_token.email, provider='google', customer=customer,
                          orgs=[domain], email=id_token.email, email_verified=id_token.email_verified)
     return jsonify(token=token.tokenize)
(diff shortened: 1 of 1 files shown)


satterly:
Thanks for the PR. I prefer your suggestion of using the first part of the email address, ie. without the domain. If you can make that small change I'll merge the PR.

icy:
You're welcome @satterly . I have updated the PR as you suggested.

alerta/auth/google.py
```diff
@@ -47,6 +47,7 @@ def google():
 
     customer = get_customer(id_token.email, groups=[domain])
 
-    token = create_token(id_token.subject, profile['name'], id_token.email, provider='google', customer=customer,
+    username = profile.get('name', id_token.email.split('@')[0])
```
satterly: This is very minor, but can you call this variable `name` not `username`.

Calling it `username` is a bit confusing because the `login` parameter (which is next in the list) is assigned to `preferred_username` within the `create_token()` method. Thanks.

kyanh-huynh: Agreed :) I will fix it shortly

satterly:
Thanks! 👍

PR #418, opened 2017-11-09
Recieve telegram webhook config from env

freeseacher:
To be in sync with sender we have to take config from env and then from config 
https://github.com/alerta/alerta-contrib/blob/master/plugins/telegram/alerta_telegram.py


satterly:
Decided to keep webhook config located in the relevant local file to keep it all together, like plugins.

PR #419, opened 2017-11-10
Telegram fix settings 

freeseacher:


satterly:
Thanks for this. Can you just explain why you need the Telegram webhook settings to be available as environment variables and can't just add them to the server configuration file?

My concern is that the list of supported environment variables will become unmanageable if every possible server setting is handled this way.

freeseacher:
Cause i want to use alerta as docker container. 
There are no perfect way to config docker container with file. So standart one is to do it with env vars.

freeseacher:
> My concern is that the list of supported environment variables will become unmanageable if every possible server setting is handled this way.

I can update docs if it is eligible

satterly:
I don't understand the challenges with deploying Docker in production environments well enough as we don't use Docker at work. What orchestration tool do you use to deploy your docker containers? Kubernetes, Docker Swarm, other? Haven't any of them solved the configuration problem satisfactorily yet?

freeseacher:
i do not use orchestration tool. just docker-compose. 
it is very useful when all config done only via one file -- docker-compose.yml

satterly:
Docker entrypoint scripts can deal with local config files. Either way, I've decided to accept [the first version](alerta/alerta#418) of this pull request after all. For now it probably makes sense to keep the config for a webhook located in the webhook file.

PR #435, opened 2017-12-11
Graylog webhook

tester22:
An webhook to received alerts from Graylog: http://docs.graylog.org/en/2.3/pages/streams/alerts.html#http-alert-notification

Note: Graylog doesn't support many parameters but the url is defined for each alert.
Therefor it is possible to add most parameters as url params:
http://api.alerta.io/webhooks/graylog?environment=Production&event=404errors&service=NGINX,Reverseproxy

The current available parameters are:
event
event_type
environment
service
severity


alerta/webhooks/graylog.py
```diff
@@ -0,0 +1,159 @@
...
+#                     "backlog": 5,
+#                     "time": 1
+#                 }
+#             }
+#         ],
+#         "id": "55f1509dbee8e2841898eb64",
+#         "title": "test",
+#         "content_pack": null
+#     }
+# }
```
satterly: Can you move the example payload to https://github.com/alerta/alerta/blob/master/tests/test_webhooks.py and add a test?

satterly:
Looks good but I can't be sure it works unless there's a test case. Will merge once you've added one. Thanks!

satterly:
Good work. Thanks! 👍

PR #441, opened 2017-12-26
Simplify model imports

satterly:


satterly:
The 'models' need to be lazy loaded otherwise there are circular imports during app startup.

PR #444, opened 2018-01-08
Timeout need to be in ms

tester22:
The timout need to be multiplied with 1000 to be in ms instead of seconds.


satterly:
Good find. You can see that it used to be `1000` when it was a cron job... https://github.com/alerta/alerta/blob/release-4.10.x/contrib/mongo/housekeepingAlerts.js#L8

PR #450, opened 2018-01-22
Housekeeping Endpoint and setup.py

mmulqueen:
I've reworked /management/housekeeping to send JSON responses. See also alerta/python-alerta-client/pull/118

I've also put POST as an allowed method. I think that POST is more correct than GET in this case. This is a minor concern because anyone calling this endpoint is likely to be in full control of their environment. I've left GET there for backwards compatibility.

I've also made some minor modifications to setup.py because I was seeing failures installing from the sdist locally.

All of this tested with Python 3.5 on Debian 9.

### DIFF
--- a/alerta/management/views.py
+++ b/alerta/management/views.py
@@ -128,7 +128,7 @@ def health_check():
     return 'OK'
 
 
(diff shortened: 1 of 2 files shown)


satterly:
Good work. Thank you. 👍

PR #455, opened 2018-01-25
Dynamically create customer-related queries

satterly:
Should be much more efficient to dynamically build the query if customer views is being used than relying on "param IS NULL or col1=param" pattern. see https://stackoverflow.com/questions/23386078/how-to-remove-conditions-from-where-clause-if-parameters-are-null


satterly:
Fixes #458

PR #482, opened 2018-03-02
Fix segregation of blackouts in customer views

middelthun:
Blackouts in customer views are not segregated, meaning one customer's blackout affects any customer's alerts.

This fix segregates the blackouts in customer views, and defines that customer==NULL is a global blackout affecting all customers.


satterly:
Thanks for the PR. Sorry it took so long to merge but I thought I should write some tests for this first (#498). Good work. 👍

PR #486, opened 2018-03-12
Fix: blackout for one customer blackouts all customers

arindamchoudhury:
Now, if we create blackout for one customer, it blackouts all customers. this pull request fixes it.


satterly:
Thanks for the PR but this didn't take account of global blackouts created by admins so I will close this one in favour of #482 which did take them into account.

PR #490, opened 2018-03-15
Support regex in ALLOWED_ENVIRONMENTS config setting.

nabadger:
This allows alerta to deal wtih environments in a more flexible manner, without needing to make
configuration changes when an environment is modifiied.

Our environments may end up being cluster specific, so by enabling regex support, we can do something like:

```
# Actual environments may be:
# - dev-cluster-eu
# - dev-cluster-us
# - prod-cluster-eu 
# - ...

ALLOWED_ENVIRONMENTS = [ "dev-cluster.*",  "prod-cluster.*" ]
```


satterly:
Thanks! 👍

PR #494, opened 2018-03-21
changes to support sso via pingfederate

sushovan23:


satterly:
LGTM but I don't actually have any knowledge of pingfederate and have no way to test this so I'm going to merge this on the basis that this has been tested thoroughly (and will be supported by) yourself. 👍

PR #495, opened 2018-03-29
Add alert value to history log if change on de-dup event

### LINKED ISSUE #483: Get History of "value" and "text" even no status/severity/event changes
Hi,

I need to have an history of value/text for duplicated alerts to have an history of value/text  over the time even no changes on the status/severity or event.

I'm not sure it's possible now but it could be interesting.

Regards

satterly:
Fixes #483


bennneuh:
I confirm, the issue seems resolved! Thanks a lot

PR #496, opened 2018-03-30
Add support for alert status of blackout

satterly:
If `NOTIFICATION_BLACKOUT` is set to `True` then alerts are processed by Alerta but the status is set to "blackout" and all plugins are skipped to prevent notifications.


3IWOH:
Thanks much for the work on this. So far everything seems to be working fine. One thing I am considering is limiting the actions that can be done on a blackout status alert. Right now someone can open and ack, etc which we don't want for blackout alerts. Looks like all of that can be done in the UI, though.

3IWOH:
I did find an issue related to this. I have set BLACKOUT_ACCEPT=['normal', 'ok', 'cleared']
This still works fine clearing the blackout alerts in alerta however I am noticing that these clearing events are still triggering a write to influx via the influxdb plugin. This is causing an issue because there is no corresponding open event (since the open blackout events do not trigger additional plugins). I have not figured out how to work around this yet...possibly using the routing plugin would be the way to go to ignore cleared blackout alerts.

PR #512, opened 2018-04-11
Support multiple OAuth2 clients for Github

satterly:
Github does not allow multiple callback URLs for a single OAuth2 application so it is necessary to create one each for web and CLI and then add them (in order) to the server config file as a comma-separated list.

**Example**
```
# Github (web,cli)
OAUTH2_CLIENT_ID = '1a725e0c58e45example,e17d26b970fd0example'
OAUTH2_CLIENT_SECRET = '2daddc40ebe236d337d80bf07fbb40e37example,dd00393551927c7e0c04d7d3acb993d76example'
```

This is only necessary if users want to use the `alerta` CLI and authenticate using Github credentials. See alerta/python-alerta-client#128

### DIFF
--- a/alerta/auth/github.py
+++ b/alerta/auth/github.py
@@ -19,10 +19,15 @@ def github():
         access_token_url = 'https://github.com/login/oauth/access_token'
         github_api_url = 'https://api.github.com'
 
+    client_lookup = dict(zip(
+        current_app.config['OAUTH2_CLIENT_ID'].split(','),
+        current_app.config['OAUTH2_CLIENT_SECRET'].split(',')
+    ))
+    client_secret = client_lookup.get(request.json['clientId'], None)
     params = {
         'client_id': request.json['clientId'],
         'redirect_uri': request.json['redirectUri'],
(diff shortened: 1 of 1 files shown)


satterly:
**Update**

To get this to work with Alerta version 6.x onwards the local configuration needs to overridden to define the OAuth2 client id for the specific client (ie. web or CLI) ...

local web config:
```$ cat app/config.json
{"endpoint": "http://api.local.alerta.io:8080", "client_id": "4f846ebb21eec62ebdxx"}
```

local CLI config:
```
$ cat ~/.alerta.conf
[DEFAULT]
profile = development

[profile development]
endpoint = http://localhost:8080
client_id = 4e70bcfea6e9535cedxx
```

server config:
```
$ grep OAUTH ~/.alertad.conf
OAUTH2_CLIENT_ID='4f846ebb21eec62ebdxx,4e70bcfea6e9535cedxx'  # web,cli
OAUTH2_CLIENT_SECRET='14ed655bd70186ca03c442eaa805fbed0ca9xxxx,444c1ce9bd07b50bb74801b9a638818fe01cxxxx'
```

PR #520, opened 2018-04-24
Test for operator close should set severity to normal

satterly:
This test will fail until the issue is fixed.

### DIFF
--- a/tests/test_alerts.py
+++ b/tests/test_alerts.py
@@ -281,6 +281,45 @@ class AlertTestCase(unittest.TestCase):
         data = json.loads(response.data.decode('utf-8'))
         self.assertEqual(data['alert']['status'], 'open')
 
+    def test_reopen_alerts(self):
+
+        # severity == warning -> status=open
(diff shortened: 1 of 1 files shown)


tests/test_alerts.py
```diff
@@ -281,6 +281,45 @@ def test_expired_alerts(self):
...
+        data = json.loads(response.data.decode('utf-8'))
+        self.assertEqual(data['alert']['status'], 'closed')
+
+        # severity == warning -> status=open
+        response = self.client.post('/alert', data=json.dumps(self.warn_alert), headers=self.headers)
+        self.assertEqual(response.status_code, 201)
+        data = json.loads(response.data.decode('utf-8'))
+        self.assertEqual(data['alert']['status'], 'open')
+        self.assertEqual(data['alert']['severity'], self.app.config['DEFAULT_NORMAL_SEVERITY'])
+        self.assertEqual(data['alert']['duplicateCount'], 0)
```
satterly: currently severity is "warning" and duplicate count is "1" when it should be "normal" and "0".

Issue #194, opened 2016-05-05
Adding LDAP authorization support to Alerta


AliGouta:
Is it on the roadmap to add support to LDAP ?

satterly:
No. The web UI uses https://github.com/sahat/satellizer for client-side authentication and that library does not support LDAP only OAuth and Basic Auth. Apparently you can get [LDAP-OAuth bridges](https://www.google.co.uk/?#q=oauth%20ldap%20bridge) that might do what you want.

PR #524, opened 2018-04-26
First implementation attempt for supporting LDAP authentication

Anderen2:
Hi!
This is a first attempt on implementing support for LDAP authentication as an alternative to the built-in base authentication. 
Note that the built-in user authorization (via user.py) is still used.

This implementation will not change anything for those already using the basic authentication, however if the configuration value "LDAP_SERVER" is specified it will use basic_ldap.py in favor of basic.py. 

It supports multiple LDAP domains/basedn's, chosen from the users domain. 

Sample config for OpenLDAP (Active Directory should also be supported, but untested.):
```
LDAP_SERVER = 'ldaps://ldap.example.com'
LDAP_DOMAINS = {
    'example.com': 'cn=%s,dc=example,dc=com'
}
```

An earlier issue on this exists (#194 ) and was closed as satellizer did not support LDAP, however this pull request does not require any frontend change to support LDAP. 

Suggestions for improvement is highly appreciated as this is my first contribution (so far) towards this project.

### DIFF
--- a/alerta/app.py
+++ b/alerta/app.py
@@ -11,6 +11,7 @@ from alerta.utils.plugin import Plugins
 from alerta.utils.config import Config
 from alerta.utils.key import ApiKeyHelper
 from alerta.utils.webhook import CustomWebhooks
+from alerta import auth
 
 config = Config()
 severity = Severity()
@@ -44,6 +45,8 @@ def create_app(config_override=None, environment=None):
     qb.init_app(app)
     sentry.init_app(app)
 
+    auth.init_app(app)
+
     from alerta.utils.format import CustomJSONEncoder
     app.json_encoder = CustomJSONEncoder
 
@@ -53,8 +56,8 @@ def create_app(config_override=None, environment=None):
     from alerta.webhooks import webhooks
     app.register_blueprint(webhooks)
 
-    from alerta.auth import auth
-    app.register_blueprint(auth)
+    from alerta.auth import auth as auth_blueprint
+    app.register_blueprint(auth_blueprint)
 
     from alerta.management import mgmt
     app.register_blueprint(mgmt)
--- a/alerta/auth/__init__.py
+++ b/alerta/auth/__init__.py
@@ -1,12 +1,17 @@
-
 from flask import Blueprint, request
 
 from alerta.exceptions import ApiError
 
-auth = Blueprint('auth', __name__)
+auth = Blueprint('auth', __name__) 
 
-from . import basic, github, gitlab, google, keycloak, pingfederate, saml2, userinfo
-
+def init_app(app):
+    from . import github, gitlab, google, keycloak, pingfederate, saml2, userinfo
+    
+    # If LDAP_SERVER is defined in config then use basic_ldap instead of basic to provide LDAP authentication
+    if "LDAP_SERVER" in app.config:
+        from . import basic_ldap
+    else:
+        from . import basic
 
 @auth.before_request
 def only_json():
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -1,0 +1,72 @@
+
+import logging
+from uuid import uuid4
+
+from flask import current_app, request, jsonify, render_template
+from flask_cors import cross_origin
+
+from alerta.auth.utils import is_authorized, create_token, get_customers
+from alerta.exceptions import ApiError
+from alerta.models.user import User
+from alerta.utils.api import absolute_url
+from . import auth
+
+import ldap
+
+@auth.route('/auth/signup', methods=['OPTIONS', 'POST'])
+@cross_origin(supports_credentials=True)
+def signup():
+    raise NotImplementedError
+
+@auth.route('/auth/login', methods=['OPTIONS', 'POST'])
+@cross_origin(supports_credentials=True)
+def login():
+    # Retrieve required fields from client request
+    try:
+        email = request.json.get('username', None) or request.json['email']
+        password = request.json['password']
+    except KeyError:
+        raise ApiError("must supply 'username' and 'password'", 401)
+
+    username = email.split("@")[0]
+    domain = email.split("@")[1]
+
+    # Validate LDAP domain
+    if not 'LDAP_DOMAINS' in current_app.config:
+        raise ApiError("LDAP_DOMAINS not configured", 500)
+
+    if domain not in current_app.config["LDAP_DOMAINS"]:
+        raise ApiError("unauthorized domain", 403)
+
+    userdn = current_app.config["LDAP_DOMAINS"][domain] % username
+
+    # Attempt LDAP AUTH
+    try:
+        ldap_connection = ldap.initialize(current_app.config['LDAP_SERVER'])
+        ldap_connection.simple_bind_s(userdn, password)
+
+    except ldap.INVALID_CREDENTIALS:
+        raise ApiError("invalid username or password", 401)
+
+    except:
+        raise ApiError("logon failed", 500)
+
+    # Create user if not yet there
+    user = User.find_by_email(email=email)
+    if not user:
+        user = User(username, email, "", ["user"], "LDAP user")
+        user.create()
+        user.set_email_verified()
+
+    # Check user is active
+    if user.status != 'active':
+        raise ApiError('user not active', 403)
+
+    # Assign customers & update last login time
+    customers = get_customers(user.email, groups=[user.domain])
+    user.update_last_login()
+
+    # Generate token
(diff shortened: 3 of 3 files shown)


alerta/app.py
```diff
@@ -44,6 +45,8 @@ def create_app(config_override=None, environment=None):
     qb.init_app(app)
     sentry.init_app(app)
 
+    auth.init_app(app)
```
Anderen2: Unsure if this is the best way, however I had issues with reading the configuration any other way. 
Issue occurs as the Flask app configuration seems not to be available at import-time.

alerta/auth/__init__.py
```diff
@@ -1,14 +1,19 @@
...
 from flask import Blueprint, request
 
 from alerta.exceptions import ApiError
 
-auth = Blueprint('auth', __name__)
-
-from . import basic, github, gitlab, google, keycloak, pingfederate, saml2, userinfo
+auth = Blueprint('auth', __name__) 
 
+def init_app(app):
```
Anderen2: As with change in alerta/app.py: 
Unsure if this is the best way, however I had issues with reading the configuration any other way.
Issue occurs as the Flask app configuration seems not to be available at import-time.

alerta/auth/basic_ldap.py
```diff
@@ -0,0 +1,72 @@
+
```
Anderen2: This file is more or less a stripped down version of alerta/auth/base.py. 
Email-verification, password-hash checking and signup have been removed.

alerta/auth/basic_ldap.py
```diff
@@ -0,0 +1,72 @@
...
+from alerta.models.user import User
+from alerta.utils.api import absolute_url
+from . import auth
+
+import ldap
+
+@auth.route('/auth/signup', methods=['OPTIONS', 'POST'])
+@cross_origin(supports_credentials=True)
+def signup():
+    raise NotImplementedError
```
Anderen2: This might not be ideal, however there is no good way to handle signup here yet without a change in https://github.com/alerta/angular-alerta-webui/

satterly:
Thanks for this. Any idea what's the easiest way to get a dev LDAP server running that I could test this agains?

alerta/auth/basic_ldap.py
```diff
@@ -0,0 +1,72 @@
...
+        raise ApiError("invalid username or password", 401)
+
+    except:
+        raise ApiError("logon failed", 500)
+
+    # Create user if not yet there
+    user = User.find_by_email(email=email)
+    if not user:
+        user = User(username, email, "", ["user"], "LDAP user")
+        user.create()
```
satterly: Why are you creating the user in the Alerta database when it already exists in the LDAP database? What purpose does that serve?

Anderen2: Hi!

This is to keep using Alerta's framework for authorization, only the authentication is offloaded to LDAP. 

Users, permissions and roles are still controlled by Alerta here. 

By the way, a colleague of mine will get back to you regarding the question regarding dev LDAP server.

satterly: I would have thought it would make sense to make full use of the LDAP database and not just use it for authentication. The LDAP server could also be used to set user roles by mapping LDAP groups to Alerta roles. Similar to how it done for the Keycloak and SAML integrations.

See these examples of how (unfortunately complicated) it can be done... http://docs.grafana.org/installation/ldap/ and https://access.redhat.com/documentation/en-us/red_hat_jboss_operations_network/3.1/html/admin_initial_setup_inventory_groups_and_users/about-ldap-groups

As a first iteration, it might be good enough to just use it for authentication and add further integration at a later date.

satterly:
I had a go at refactoring this PR to make the switch between basic and LDAP a bit more seamless ... See https://github.com/alerta/alerta/compare/ldap-refactor?expand=1

alerta/auth/basic_ldap.py
Anderen2: Hmm, I agree. 
For our usecase it might be that we require LDAP groups as well, I need to check up on that. It's very possible that you'll get another pull request for implementing LDAP groups too in the very near future

Anderen2:
Hi @satterly!
Sorry for the very slow response. I've received some instructions for how to test this with OpenLDAP on a RHEL-based system now. I would guess that testing it with a Debian-based system is not much different. 

Following are some steps to get it up and running:
```
yum -y install openldap openldap-servers openldap-clients

# Add the Manager password '1234'
echo 'olcRootPW: {SSHA}1iBZTN8V22GieckRXHVWCjgSfZ5ZyxCR' >> /etc/openldap/slapd.d/cn\=config/olcDatabase\=\{2\}hdb.ldif
Systemctl start slapd.service
systemctl enable slapd.service
ldapadd -Y EXTERNAL -H ldapi:/// -f /etc/openldap/schema/cosine.ldif
ldapadd -Y EXTERNAL -H ldapi:/// -f /etc/openldap/schema/nis.ldif
ldapadd -Y EXTERNAL -H ldapi:/// -f /etc/openldap/schema/inetorgperson.ldif

# Adding user1. Prompted for Manager password
ldapadd -vvvv -D cn=Manager,dc=my-domain,dc=com -h localhost -p 389 -f /tmp/user1.ldif  -W

# Adding password 'secret1' to user user1
ldappasswd -vvvv -s secret1 -D cn=Manager,dc=my-domain,dc=com -h localhost -p 389 -x "uid=user1,ou=users,dc=my-domain,dc=com" -W
```

The LDIF file referenced contains the following:
```
dn: dc=my-domain,dc=com
objectclass: dcObject
objectClass: organization
o: my-domain

dn: ou=users,dc=my-domain,dc=com
objectClass: organizationalUnit
ou: users

dn: uid=user1,ou=users,dc=my-domain,dc=com
objectClass: inetOrgPerson
description: user1
cn: user1
sn: user1
```

When Alerta then is configured to use the server as LDAP_SERVER, and the LDAP_DOMAINS match the LDIF configuration (example below), you should be able to log into Alerta with username "user1" and password "secret1" 

```
LDAP_SERVER = 'ldap://ldap.my-domain.com'

LDAP_DOMAINS = {
    'my-domain.com': 'uid=%s,ou=users,dc=my-domain,dc=com'
}

```

satterly:
Thanks for the detailed instructions on how to test this. It worked! I'm going to merge this "as is" but then refactor it slightly as suggested above and to rename `LDAP_SERVER` to `LDAP_URL` to make it consistent with other settings.

One thing I had to find out the hard way was that this uses the python library called `python-ldap` not `ldap`. ie. you must run `pip install python-ldap`

carlospeon:
Hello, I've tested alerta-ldap-refactor branch and It worked fine.

Thank you!

PMDubuc:
I couldn't find anything in the documentation on how to so authentication/authorization with LDAP.  I assumed it wasn't supported until I ran into this.  Is this configuration documented somewhere?

satterly:
There's no official documentation yet but this example Docker stack might be enough to get you going ... https://github.com/alerta/docker-alerta/tree/master/examples/ldap

PMDubuc:
I have another question about this, what goes into the /var/www/html/config.js to use ldap as a provider.  Using "basic" doesn't seem to work.  This is the contents of my config.js:
```

'use strict';

angular.module('config', [])
  .constant('config', {
    'endpoint'        : "http://xx.xxx.xx.xxx:80/api",
    'provider'        : "basic", // google, github, gitlab, keycloak, pingfederate, saml2 or basic
    'colors'          : {}, // use default colors
    'severity'        : {}, // use default severity codes
    'audio'           : {} // no audio
  });

```

PMDubuc:
I'm testing this with our LDAP server and noticed that the the alerta log prints the user's password in plain text.  Is there a way to obfuscate it?

Example:
alertad: *** <ldap.ldapobject.SimpleLDAPObject object at 0x7f6541f52ef0> ldaps://ldap-test:389 - SimpleLDAPObject.simple_bind#012(('CN=testuser,OU=Users,DC=xxxx,DC=xxxx,DC=com',#012  'password',#012  None,#012  None),#012 {})

satterly:
What happens if you set `DEBUG=False` in server configuration?

PMDubuc:
I'm not sure what I might be doing wrong, but I couldn't get this implementation to authenticate with our Active Directory based LDAP or with our Linux based LDAP server even though I could get a search to work with the ldapsearch utility using the same DN and password.  I tried a few modifications to basic_ldap.py to set OPT_REFERRALS to 0 and the protocol_version to 3 as suggested [here](https://blog.thomastoye.be/python-ldap-authentication-with-microsoft-active-directory-46661bebc483) but it still didn't work.

This is just FYI, I think we'll have to settle for  basic authentication for Alerta for now.

PR #527, opened 2018-04-27
saml_client bug

d-ganchar:
I found bug with **saml2** config. How to reproduce:

alerta.conf:

```
import json

SAML2_CONFIG = json.loads('{"attribute_map_dir": "dir_path", "service": {"sp": {"allow_unsolicited": true}}}')

SAML2_CONFIG['key_file'] = 'key'
SAML2_CONFIG['cert_file'] = 'cert'
SAML2_CONFIG['disable_ssl_certificate_validation'] = True

SAML2_CONFIG['metadata'] = {'local': ['path_to_meta.xml']}
```
Set path of **alerta.conf**  to **ALERTA_SVR_CONF_FILE** env variable. Try to open **/auth/saml**

Error(I do not have trace but the reason is next): Config do not have metadata(None).

Env:
    python 3.6.1
    MacOS sierra 10.12.5

### DIFF
--- a/alerta/auth/saml2.py
+++ b/alerta/auth/saml2.py
@@ -19,13 +19,10 @@ except ImportError:
     pass  # saml2 authentication will not work
 
 
-def spConfig():
-    return saml2.config.Config()
+def get_saml2_config():
+    config = saml2.config.Config()
 
-
-def saml_client():
-
-    saml2_config_default = {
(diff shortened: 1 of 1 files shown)


satterly:
Thanks. I'll assume this is correct. We should probably have tests for auth endpoints though. 👍

Issue #528, opened 2018-04-27
Housekeeping fails when using postgresql


Anderen2:
It seems like the latest update implementing shelving of alarms broke the housekeeping for PostgreSQL ( https://github.com/alerta/alerta/commit/0d36ac624a481e01c5d0598603582824219c578b ).

Sample housekeeping call:
```
[root@snip ~]# curl -H "Authorization: Key snip" localhost/api/management/housekeeping 
{
  "code": 503, 
  "errors": null, 
  "message": "HOUSEKEEPING FAILED: Type names and field names can only contain alphanumeric characters and underscores: '?column?'", 
  "status": "error"
}
```

UWSGI log:
```
2018-04-28 00:06:43,862 - alerta[18702]: ERROR - HOUSEKEEPING FAILED: Type names and field names can only contain alphanumeric characters and underscores: '?column?' [in /usr/lib/python2.7/site-packages/alerta_server-5.2.0_-py2.7.egg/alerta/exceptions.py:67]
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/lib/python2.7/site-packages/flask_cors/decorator.py", line 128, in wrapped_function
    resp = make_response(f(*args, **kwargs))
  File "/usr/lib/python2.7/site-packages/alerta_server-5.2.0_-py2.7.egg/alerta/auth/utils.py", line 95, in wrapped
    return f(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/alerta_server-5.2.0_-py2.7.egg/alerta/management/views.py", line 148, in housekeeping
    raise ApiError('HOUSEKEEPING FAILED: %s' % e, 503)
ApiError: HOUSEKEEPING FAILED: Type names and field names can only contain alphanumeric characters and underscores: '?column?'
```

PostgreSQL version and query output
```
[root@snip ~]# psql -V    
psql (PostgreSQL) 10.1

alerta=> SELECT id, event, 'expired', last_receive_id                                                                                                                                                            
              FROM alerts
             WHERE status NOT IN ('expired','shelved') AND timeout!=0
               AND (last_receive_time + INTERVAL '1 second' * timeout) < (NOW() at time zone 'utc')
;
[...]

PR #529, opened 2018-04-29
Fix for issue where housekeeping has stopped working with PostgreSQL

Anderen2:
This fixes issue #528 where housekeeping fails when using PostgreSQL after the new shelving feature.

### DIFF
--- a/alerta/models/alert.py
+++ b/alerta/models/alert.py
@@ -468,19 +468,26 @@ class Alert(object):
 
     @staticmethod
     def housekeeping(expired_threshold=2, info_threshold=12):
-        for (id, event, status, last_receive_id) in db.housekeeping(expired_threshold, info_threshold):
-            if status == 'open':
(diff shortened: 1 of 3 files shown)


alerta/models/alert.py
```diff
@@ -468,19 +468,26 @@ def get_tags(query=None):
 
     @staticmethod
     def housekeeping(expired_threshold=2, info_threshold=12):
-        for (id, event, status, last_receive_id) in db.housekeeping(expired_threshold, info_threshold):
-            if status == 'open':
-                text = 'unshelved after timeout'
-            elif status == 'expired':
-                text = 'expired after timeout'
-            else:
-                text = 'alert timeout status change'
```
Anderen2: Please note that I could not see any codepath where text would be set to 'alert timeout status change'. It is therefore omitted from my pull request.

Issue #534, opened 2018-05-08
improvement: Requirements modifications


d-ganchar:
At the moment requirements in [setup.py](https://github.com/alerta/alerta/blob/master/setup.py#L39) and in [requirements.txt](https://github.com/alerta/alerta/blob/master/requirements.txt) are different.
Need to add parsing of **requirements.txt** in **setup.py** and use 1 list for installing dependencies.
Also need to freeze versions of packages(such as Flask, requests, PyJWT etc.).
What do you think?

d-ganchar:
I propose to parse **requirements.txt** in **setup.py**. What is benefits:

- We will have 1 file with dependencies(requirements.txt)
- Less opportunity for error

**About freezing**. I propose set min versions of support. For example, at now we have:

```
pytz
PyJWT
bcrypt
```

But what is the version? We should to set available range of versions like **package>=0.2,<0.3**.

satterly:
When you said "freeze versions" I thought you meant only support one version of a library. I would agree setting a minimum version or a range would be useful.

satterly:
Looks like the recommended approach is to not ship a `requirements.txt` at all, but if you really need to then it should literally just be a full stop `.`. See https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py

PR #536, opened 2018-05-10
#534 dependencies

d-ganchar:
Added min range of versions.
**setup.py** uses _requirements.txt_ dependencies.


.gitignore
```diff
@@ -8,6 +8,7 @@ ve
 dist
 build
 .idea
+env
```
d-ganchar: do not track virtualenv folder

satterly: I'm not sure why your virtualenv folder is in the source code folder, and I don't think you can always assume someone is going to call this folder `env`.

requirements.txt
```diff
@@ -1,11 +1,11 @@
...
-pytz
-PyJWT
-bcrypt
+pymongo>=3.6,<3.7
+psycopg2>=2.7,<2.8
+requests>=2.18,<2.19
+python-dateutil>=2.7,<2.8
+pytz==2018.4
+PyJWT>=1.6,<1.7
+bcrypt>=3.1,<3.2
```
satterly: What is the rational behind choosing these particular version ranges?

satterly:
Thanks for your PR but the simpler solution is just to replace the `requirements.txt` with a dot ".".

.gitignore
d-ganchar: It is very convenient when `env` folder is in project. 
By the way, the folders `env`,` venv`, `ENV` are always added to gitignore, if you use auto-generation of gitignore for `python` projects. So I think it's make sense because a lot of developers creates `env` folder inside project.

requirements.txt
d-ganchar: I decided that it's better to use the latest versions, because they are often more stable and with fewer bugs. Also I'm sure that only changes in the `patch` version can guarantee backward compatibility. `major` and `minor` changes can just break something. 
What do you think?

.gitignore
satterly: See #540

PR #548, opened 2018-05-23
Allow users with a customer domain mapping to '*' to view all customers

Anderen2:
Hi @satterly !

This pull request allows users with a wildcard domain mapping to view all customers (as those specified in "ADMIN_USERS" in the configuration may). 

When specifying a single "*" in the customer field, all users part of that domain will be able to view alarms for all customers.

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -1027,8 +1027,13 @@ class Backend(Database):
             response = g.db.customers.find_one({"match": match}, projection={"customer": 1, "_id": 0})
             if response:
                 customers.append(response['customer'])
+
         if customers:
+            if '*' in customers:
+                return '*' # all customers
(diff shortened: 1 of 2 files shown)


Anderen2:
Adding a usecase here too in case the intent is unclear. 

We are an managed service provider with very many different customers, and we have some hundred technicians that should be able to view alarms for all customers. 

Not all of these technicians should have admin access to Alerta. 
Neither is updating the "ADMIN_USERS" list for all new and leaving users in a configuration on a cluster a very pretty task. 

As far as I see there is no way other than using "ADMIN_USERS" to view all alarms, this pull request implements one based on the users domain. 

If there already exists, or if there is a better way to implement this then please say so, and I'll modify my pull-request to match!

satterly:
Good idea. Thanks! 👍

PR #551, opened 2018-05-25
webhooks: grafana alert property override with webhook query parameters

melor:
Grafana webhook alert properties can be controlled by overriding the
default ones with webhook url parameters:

Example (split on multiple lines for readability):

  https://alerta/webhooks/grafana?api-key=xxx
      &environment=Production
      &event_type=performanceAlert
      &group=Performance
      &origin=Grafana
      &service=Grafana
      &severity=major


satterly:
Thanks! 👍

Issue #552, opened 2018-05-29
Plugin "status_change" not triggered by actions done in WebUI


Anderen2:
Hi!

It seems like plugins/integrations using the "status_change" hook does not trigger when actions are performed from the WebUI or via the actions API (/alert/<alert_id>/action), only when a status is explicitly set via the API (/alert/<alert_id>/status).

It seems like this was something that changed with https://github.com/alerta/alerta/pull/521 when the "process_action" function was added ( https://github.com/alerta/alerta/blame/master/alerta/utils/api.py#L147 )

Is this working as intended, or would you accept me writing a pull request for enabling "plugin.status_change" to be triggered also in process_action ( https://github.com/alerta/alerta/blob/master/alerta/utils/api.py#L147 )?

satterly:
This was the initial intent but if there's a use case for changing it then we can review it. What is your specific use-case?

Anderen2:
We use a plugin to integrate Alerta with our ticketing system (via AMQP), it works great for everything except changes done from the WebUI in Alerta. 

When the alarm is closed manually in Alerta we never receive the clear, nor closed state, making "stale" tickets and leads to confusion on how / why the alarm no longer exists in Alerta (especially if housekeeping has done it's job too).

unk1nd:
If this was the initial intent, then i think that it should be fixed

Anderen2:
An additional usecase example we face is being able to keep two monitoring systems in sync. 

We have some monitoring done with SCOM for some systems, we are currently shipping all alerts, and updates done in SCOM towards Alerta as we'd like to use Alerta as the primary and unified tool. 

However, closing SCOM alerts in Alerta is then not possible, as the alarm is still open in SCOM and only gets re-raised to Alerta. 

If actions done towards alerts would trigger the status_change hook for plugins/integrations we would be able to close alerts in SCOM when they were closed in Alerta.

PR #555, opened 2018-06-07
Implementation of triggering plugin status_change on actions (#552)

Anderen2:
This pull request implements triggering of "status_change()" for plugins on actions as discussed in issue #552


satterly:
Thanks for the contribution but it needed considerably more refactoring for it to work correctly. See #556

Anderen2:
@satterly Thanks a lot!

PR #571, opened 2018-07-11
Get plain body if json and text is missing

tester22:
This PR makes it possible to use inbound webhooks with type application/x-www-form-urlencoded.


satterly:
Excellent. Thanks! 👍

Issue #570, opened 2018-07-11
Mongo style query using 'q' parameter doesn't work


mehertz:
Hi,

I've been trying to use the `q` parameter as described here:
http://alerta.readthedocs.io/en/latest/api/reference.html#search-alerts

Unfortunately it doesn't seem to work. I get the following error (reqeust is `/api/alerts?q=awd`):
```
{
  "code": 500, 
  "errors": [
    "Traceback (most recent call last):\n  File \"/usr/lib/python2.7/site-packages/flask/app.py\", line 1612, in full_dispatch_request\n    rv = self.dispatch_request()\n  File \"/usr/lib/python2.7/site-packages/flask/app.py\", line 1598, in dispatch_request\n    return self.view_functions[rule.endpoint](**req.view_args)\n  File \"/usr/lib/python2.7/site-packages/flask_cors/decorator.py\", line 128, in wrapped_function\n    resp = make_response(f(*args, **kwargs))\n  File \"/usr/lib/python2.7/site-packages/alerta/auth/utils.py\", line 118, in wrapped\n    return f(*args, **kwargs)\n  File \"/usr/lib/python2.7/site-packages/alerta/models/metrics.py\", line 259, in wrapped\n    response = f(*args, **kwargs)\n  File \"/usr/lib/python2.7/site-packages/alerta/utils/api.py\", line 28, in decorated\n    return func(*args, **kwargs)\n  File \"/usr/lib/python2.7/site-packages/alerta/views/alerts.py\", line 199, in search_alerts\n    query = qb.from_params(request.args, query_time)\n  File \"/usr/lib/python2.7/site-packages/alerta/database/backends/mongodb/utils.py\", line 23, in from_params\n    query = json.loads(params.pop('q'))\n  File \"/usr/lib/python2.7/site-packages/werkzeug/datastructures.py\", line 179, in pop\n    is_immutable(self)\n  File \"/usr/lib/python2.7/site-packages/werkzeug/datastructures.py\", line 29, in is_immutable\n    raise TypeError('%r objects are immutable' % self.__class__.__name__)\nTypeError: 'ImmutableMultiDict' objects are immutable\n"
  ], 
  "message": "'ImmutableMultiDict' objects are immutable", 
  "status": "error"
}
```

The line where the exception is thrown is:
https://github.com/alerta/alerta/blob/master/alerta/database/backends/mongodb/utils.py#L23

[...]

PR #573, opened 2018-07-12
Fix issue whereby mongo style queries do not work

mehertz:
This is in reference to https://github.com/alerta/alerta/issues/570

params is immuteable, so we shouldn't pop from it.


satterly:
Thanks. And bonus points for writing a test! 👍 💯

PR #575, opened 2018-07-12
Add setting for non-fatal index creation handling

### LINKED ISSUE #569: Fail to create MongoDB indexes on Azure
>I have an problem when using Alerta with Azure Cosmos MongoDB, when creating the indexes I get an exception from the server causing Alerta to fail. I solved this by changeing the following lines and creating the indexes manually: https://github.com/tester22/alerta/blob/90d1343757f285c346b44b19245d7e564f7ca3dc/alerta/database/backends/mongodb/base.py#L36 I know this isn't the cleanest solution but would you consider merging it?
/cc @tester22

satterly:
Set `MONGO_RAISE_ON_ERROR=False` if creating MongoDB indexes manually and Alerta API should not fail to start if automatic index creation fails.

Fixes #569

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -28,6 +28,17 @@ class Backend(Database):
             db = self.client.get_default_database()
 
         # create unique indexes
+        try:
+            self._create_indexes(db)
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


tester22:
Yep, looks good.

satterly:
Thanks for the quick review. 😄

PR #576, opened 2018-07-17
Add option line to accept connections from ldaps

francopeapea:
Added set_option line to accept connections from ldaps://

### DIFF
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -32,6 +32,7 @@ def login():
     # Attempt LDAP AUTH
     try:
         trace_level = 2 if current_app.debug else 0
+        ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_NEVER)
         ldap_connection = ldap.initialize(current_app.config['LDAP_URL'], trace_level=trace_level)
         ldap_connection.simple_bind_s(userdn, password)
     except ldap.INVALID_CREDENTIALS:


satterly:
Can you add an explanation of what those settings mean and do? Thanks.

satterly:
Can you provide more info on what the options do and if they prevent connections from non-https LDAP?

Anderen2:
We currently use ldaps without the options mentioned here, simply by prefixing the host in LDAP_URL with 'ldaps://' (eg. 'ldaps://my.ldapserver.com'). 

I'm not sure what ldap.OPT_X_TLS_REQUIRE_CERT and ldap.OPT_X_TLS_NEVER will change, could you clarify @francopeapea ?

mmulqueen:
I just want to chime in against this PR because this reduces the security of the LDAP integration.

Essentially what this does is it doesn't bother to ask for a certificate from the LDAP server and blindly trusts the LDAP host it has connected to. I believe you still get privacy (encryption), but you lose authentication (validated certificate).

The default setting in python-ldap is OPT_X_TLS_DEMAND and OPT_X_TLS_HARD, this requires a valid certificate to be given and fails if it's not. This should remain the default for alerta.

OPT_X_TLS_NEVER is basically a work around for a self-signed certificate but it's a bad idea. Everyone should be setting up their own self-signed CA or using something like LetsEncrypt or a paid certificate. It definitely shouldn't be the default for everyone. Perhaps it should be available as an option for when people need to talk to LDAP servers outside of their control that use self-signed certificates.

satterly:
I'm going to close this based on feedback received above. Any workable solution (eg. providing the ability to set additional options) will require a different approach and new pull request. Cheers.

PR #581, opened 2018-07-25
Allow Content-type other than application/json for POST requests to /…

Chris-Lamb:
…auth/saml endpoint

* Allow the IdP to send the SAMLResponse and RelayState to the Assertion Consumer Service (Alerta) via POST request with Content-Type: application/x-www-form-urlencoded

### DIFF
--- a/alerta/auth/__init__.py
+++ b/alerta/auth/__init__.py
@@ -15,5 +15,8 @@ from . import github, gitlab, google, keycloak, pingfederate, saml2, userinfo
 
 @auth.before_request
 def only_json():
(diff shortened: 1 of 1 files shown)


satterly:
Can we change the condition logic here to make the exception for SAML more explicit?

Something like ...
```
def only_json():
    if request.path == '/auth/saml':  # Note: Exception for SAMLResponse and RelayState
        return
    if request.method in ['POST', 'PUT'] and not request.is_json:
        raise ApiError("POST and PUT requests must set 'Content-type' to 'application/json'", 415)
```

PR #590, opened 2018-08-01
Add support for Basic Auth

satterly:


alerta/auth/decorators.py
```diff
@@ -0,0 +1,102 @@
...
+                    username, password = base64.b64decode(credentials).decode('utf-8').split(':')
+                except Exception as e:
+                    raise ApiError(str(e), 401)
+
+                user = User.check_credentials(username, password)
+                if not user:
+                    raise ApiError("Invalid username or password", 401)
+                if current_app.config['EMAIL_VERIFICATION'] and not user.email_verified:
+                    raise ApiError('email not verified')
+                if is_authorized('ALLOWED_EMAIL_DOMAINS', groups=[user.domain]):
```
satterly: The logic of `is_authorized()` seems to be reverse to what you would expect. Check this.

alerta/auth/decorators.py
```diff
@@ -0,0 +1,102 @@
...
+                    raise ApiError("unauthorized domain", 403)
+
+                g.user = user.email
+                g.customers = get_customers(user.email, groups=[user.domain])
```
satterly: Should `get_customers()` be a staticmethod in the Customers model?

tests/test_auth.py
```diff
@@ -224,3 +224,7 @@ def test_login(self):
         # delete customer mapping
         response = self.client.delete('/customer/' + customer_id, headers=self.headers)
         self.assertEqual(response.status_code, 200)
+
+    def test_basic_auth(self):
+
```
satterly: Add some tests.

PR #609, opened 2018-08-22
Add auditing info to blackout model

### LINKED ISSUE #585: Extend blackout data model to support auditing
> support for comments, creation time and storing the user/key which performed a blackout 

@Anderen2

satterly:
New fields added to blackouts to improve auditing: "user", "createTime" and "text" for comments.

No database changes are required for MongoDB but three columns are added during startup if they don't exist to the "blackouts" table for Postgres. See `schema.sql` file for more details.

Fixes #585

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ last to support Python 2.7. Only critical bug fixes will be backported to Releas
 
 The only mandatory dependency is MongoDB or PostgreSQL. Everything else is optional.
 
-- Postgres version 9.6 or better
+- Postgres version 9.5 or better
 - MongoDB version 3.x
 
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ addons:
     - mongodb-3.0-precise
     packages:
     - mongodb-org-server
-  postgresql: "9.5"
+  postgresql: "9.6"
```
satterly: Could still support Postgres 9.5 if we add a `POSTGRES_RAISE_ON_ERROR` setting that could be set to `False` and it would be assumed that the database is managed manually.

PR #627, opened 2018-09-04
Add blackout permutations

middelthun:
We've come to realize that the is_blackout_period check only compares against a select few combinations of the things available for a blackout, and these are not sufficient to support our daily use of Alerta.

This PR adds all the different permutations, making it possible to blackout using any available combination.

I've added the permutations in a binary order, so it should be fairly easy to follow.

Please note that the mongodb code is untested, as I have nothing to test against.


satterly:
Thanks for this. 👍 I'll add some tests to ensure MongoDB and Postgres work as expected.

Issue #623, opened 2018-09-03
externalUrl as a link in Prometheus webhook


n0guest:
Hello.

Is there a reason why "externalUrl" field from annotations isn't rendered as a link? I'm asking because in the same time field "moreInfo" [rendered as http link](https://github.com/alerta/alerta/blob/master/alerta/webhooks/prometheus.py#L63):

```python
    if external_url:
        annotations['externalUrl'] = external_url
    if 'generatorURL' in alert:
        annotations['moreInfo'] = '<a href="%s" target="_blank">Prometheus Graph</a>' % alert['generatorURL']
```

IMHO, it would be useful to have both of them rendered as a link. @satterly, wouldn't you agree?

n0guest:
From gitter chat:

>  n0guest: satterly, sorry for hijacking discussion. But could you, please, take a look at alerta/alerta#623 I just want to clarify this matter. And if you'd agree with such small changes I could send a PR (even if it's so small).
> satterly: n0guest yep, that would be great. thanks.

Please, check PR #629. 

Thank yoiu.

satterly:
Good work. Thanks again. 👍

PR #629, opened 2018-09-04
render externalUrl as external link (webhooks/prometheus)

n0guest:
Changed rendering of externalUrl as discussed in alerta/alerta#623.


satterly:
Thanks 👍

PR #631, opened 2018-09-04
Test multiple combinations for enhanced blackout

satterly:
These tests will fail until #627 is merged.


alerta/database/backends/postgres/base.py
```diff
@@ -557,14 +557,14 @@ def is_blackout_period(self, alert):
...
-              OR ( resource=%(resource)s AND service::text[] && %(service)s AND event=%(event)s AND "group"=%(group)s AND tags <@ %(tags)s )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event IS NULL AND "group" IS NULL AND tags='{}' )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event IS NULL AND "group" IS NULL AND tags <@ %(tags)s )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event IS NULL AND "group"=%(group)s AND tags='{}' )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event IS NULL AND "group"=%(group)s AND tags <@ %(tags)s )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event=%(event)s AND "group" IS NULL AND tags='{}' )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event=%(event)s AND "group" IS NULL AND tags <@ %(tags)s )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event=%(event)s AND "group"=%(group)s AND tags='{}' )
+              OR ( resource=%(resource)s AND service <@ %(service)s AND event=%(event)s AND "group"=%(group)s AND tags <@ %(tags)s )
```
satterly: The change from `AND service::text[] && %(service)s` to `AND service <@ %(service)s` brings services into line with tags so that a blackout rule will only match if ALL services listed in the rule are services associated with that alert.

For example, if a blackout rule includes a list of services called `['Core','Web']` then ...
 * an alert only with service `Core` will not match
 * an alert with services `Core` and `Network` will not match
 * an alert with services `Core` and `Web` will match and be suppressed
 * an alert with services `Core`, `Web`, `Network` will match and be suppressed

middelthun: Perfect, thank you!

PR #665, opened 2018-09-21
Do not try starttls on the smtp if smtp_user_ssl is true

### LINKED ISSUE #664: [Mail] Should not try starttls if ssl is true
Hi,

Alerta should not try starttls on the smtp if ssl is already on.

Toniob:
close #664


satterly:
Sorry. Can you rebase this against master. I had to roll back some changes. Thanks.

PR #673, opened 2018-09-26
Fix setting timeout=None incorrectly through API

infernix:
Not sure if this is the right approach, but it's definitely broken as-is, because if I submit a status change using the API without explicitly specifying the timeout, it gets set to None, and housekeeping will fail to clean these up correctly.

### DIFF
--- a/alerta/models/alert.py
+++ b/alerta/models/alert.py
@@ -357,7 +357,7 @@ class Alert:
 
     # set alert status
     def set_status(self, status, text='', timeout=None):
-        self.timeout = timeout or current_app.config['ALERT_TIMEOUT']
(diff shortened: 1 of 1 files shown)


alerta/models/alert.py
```diff
@@ -366,10 +366,10 @@ def set_status(self, status, text='', timeout=None):
             change_type='status',
             update_time=datetime.utcnow()
         )
-        return db.set_status(self.id, status, timeout, history)
+        return db.set_status(self.id, status, self.timeout, history)
```
satterly: I think it's the "self" on line 360 that needs to be removed and that's it.

infernix:
Tested, works :) guess self.timeout isn't needed/used.

PR #677, opened 2018-09-30
SMTP enhancements

Toniob:
- Use sslcontext for the smtp connection
- remove the unnecessary call to ehlo()
- check if smtp_username is set before login()
- call to quit() instead of close() to terminate the session


satterly:
Thanks for following up on #665. Much appreciated. 👍

PR #687, opened 2018-10-06
Factor out Alerta state machine into Flask extension

satterly:


alerta/models/alert.py
```diff
@@ -477,3 +476,30 @@ def housekeeping(expired_threshold=2, info_threshold=12):
...
+            status=self.status,
+            text=text,
+            change_type='action',
+            update_time=datetime.utcnow()
+        )
+        db.set_severity_and_status(self.id, severity, status, self.timeout, history)
+        return severity, status
+
+    def is_status_blackout(self):
+        return False
```
satterly: Fix this!

PR #708, opened 2018-10-12
[WIP] Trigger user-configurable actions in plugins

### LINKED ISSUE #554: [Feature Request/Idea] Create issue in configurable systems
There has been some work to create issues in Jira from the web UI. Not sure about the status of that feature but I'd really like to see a more comprehensive solution for this:

- Should be supported by the API and on top of that by all clients (CLI, web, mobile)
- The backend should allow plugins for various systems like Jira, GitLab, BitBucket, YouTrack and others
- Maybe the information about the created ticket should be linked to the alert so that we could update the issue when new related alerts arise

Is that something you would consider for a future Alerta version?

satterly:
The server config would be something like ...
```
PLUGINS=['reject', 'blackout', 'gitlab']  # add the "gitlab" plugin with "take_action()" method
ACTIONS = ['createIssue']  # add "Create Issue" action to web UI which "gitlab" plugin supports
```
screenshot 2018-10-12 at 19 05 47

Whenever a "PUT /action/:id/action" request is received all plugins with a "take_action()" method are run with "alert", "action" and "text". It is the plugins responsibility to only action things it is responsible for.

```python
    def take_action(self, alert, action, text):
        if action == 'createIssue':
            url = '{}/projects/{}/issues?title=foo'.format(GITLAB_URL, GITLAB_PROJECT_ID)
            r = requests.post(url, headers={'Private-Token': GITLAB_ACCESS_TOKEN})
            issue_id = r.json().get('id', None)
            alert.attributes['actionId'] = issue_id
        else:
            print('action ignored by gitlab action handler')
        return alert, action, text
```
Update the alert attribute "actionId" (or whatever) with relevant info and return.

screenshot 2018-10-12 at 19 10 25

Fixes #554


Anderen2:
Looks good! 
However, would it be any way here in the future to redirect the user to a different page based on a action too? 

Usecase could be to redirect, or open a new tab/window containing the configuration page for the system responsible for creating the alarm in question.

satterly:
> Looks good!
> However, would it be any way here in the future to redirect the user to a different page based on a action too?
> 
> Usecase could be to redirect, or open a new tab/window containing the configuration page for the system responsible for creating the alarm in question.

This is already possible by passing the URL to any external system as an alert attribute. It will be rendered clickable in the web UI and should open a new tab.

PR #727, opened 2018-10-24
Add proxy fix option if running behind SSL terminating proxy

### LINKED ISSUE #715: HTTP/S of Endpoint Overwritten by Server Config
My UI deployment has the config.json of `{"endpoint":"https://my_server/alerta/api"}` 

This works, and when i go to the UI i can see it makes an initial request to `https://my_server/alerta/api/config` however that call overwrites the `endpoint` to be `http://my_server/alerta/api`

This leads to a "http called from https context" error and all subsequent requests fail. Not quite managed to track-down where/why the endpoint is getting overwritten by the server - hoping someone may be able to spot it.

satterly:
Fixes #715

### DIFF
--- a/alerta/app.py
+++ b/alerta/app.py
@@ -4,6 +4,7 @@ from flask import Flask
 from flask_compress import Compress
 from flask_cors import CORS
 from raven.contrib.flask import Sentry
+from werkzeug.contrib.fixers import ProxyFix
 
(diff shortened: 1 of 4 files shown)


satterly:
[protocol change](https://trello.com/c/GNQSqLEA/483-protocol-change)

PR #735, opened 2018-10-29
changes to allow plugin methods to be used for tagging/untagging, changing attributes and deleting alerts

frankd4github:
Bulk and housekeeping actions were not touched

New process_tag(), process_untag(), process_attributes() and process_delete() created on utils/api.py similar to the existing process_action(). views/alerts.py adjusted to use new util functions


satterly:
Thanks. I'll take a look by the end of the week.

satterly:
Sorry it's taken so long to get around to reviewing this PR. After taking some time to look at it I think it is unnecessarily complex.

For example, I'm not sure why tagging an alert would trigger a plugin method that could alter the alert (in any way) that could then trigger the status_change plugin method to also run. At times it is a struggle to reason about the flow of alerts through Alerta and unfortunately I think this would only make things more difficult.

I suspect this was done because you've taken a look at the "action_alert" endpoint and just repeated the logic. The problem with that is that "action_alert" is the most complicated endpoint (for historical reasons) and so it's not the best one to provide a template of what to do.

Below, I've used your approach but simplified it considerably to what I think should happen when an alert is tagged (for example) ...

**View method**
```python
def tag_alert(alert_id):
    tags = request.json.get('tags', None)
    text = request.json.get('text', 'tagged with %s' % tags)

    if not tags:
        raise ApiError("must supply 'tags' as json list")
    
    customers = g.get('customers', None)
    alert = Alert.find_by_id(alert_id, customers)

    if not alert:
        raise ApiError('not found', 404)

    try:
        process_tag(alert, tags, text)
    except Exception as e:
        raise ApiError(str(e), 500)

    if alert.tag(tags):
        return jsonify(status='ok')
    else:
        raise ApiError('failed to tag alert', 500)
```

**API method**
```python
def process_tag(alert: Alert, tags: List[str], text: str) -> None:

    for plugin in plugins.routing(alert):
        try:
            plugin.alert_tagged(alert, tags, text)
        except NotImplementedError as e:
            pass  # plugin does not support alert_tagged() method
        except Exception as e:
            if current_app.config['PLUGINS_RAISE_ON_ERROR']:
                raise ApiError("Error while running alert_tagged plug-in '{}': {}".format(plugin.name, str(e)))
            else:
                logging.error("Error while running alert_tagged plug-in '{}': {}".format(plugin.name, str(e)))
```

However, looking at what I've left out raises a few questions about the logical consistency of what is being proposed. Particularly when it comes to handling suppressed alerts and what to do when alerts are deleted.

I'm beginning to think a different approach is required. I know you considered using triggers on Postgres and the MongoDB opslog but perhaps there is a better solution. I need to spend some more time considering the alternatives. Let me know if you have any thoughts. I'll get back to you soon.

satterly:
This looks promising... http://flask.pocoo.org/docs/1.0/signals/

frankd4github:
Nothing to be sorry about. We all in the alerta community appreciate your time. 

Looking at signals support now. It does indeed look promising. Will most probably spend some time on this over the weekend

satterly:
I had time so took a deeper look myself. Checkout PR #744 and let me know what you think. Thanks.

satterly:
Closing in favour of #744. Thanks for your efforts on this.

PR #756, opened 2018-11-16
Make is_duplicate() return count not boolean

### LINKED ISSUE #752: Plugin question: modifying displayed alert values in the post_receive() method
I would like to modify the alert.value property in the post_receive() method of the plugin I'm writing so that it shows up on the UI dashboard.  I seems that this is only possible in the pre_receive() method, before the alert is persisted to the database.  I wan't to change a value based on the duplicateCount, which only seems to be accessible in post_receive(). But, when I change the value of he alert there, the new value isn't shown on the display.   Is there any way of doing this?

satterly:


satterly:
Close. Wrong approach.

PR #758, opened 2018-11-20
Fix custom webhook view to avoid alert processing

bcanyelles:
Fix custom webhook view to avoid alert processing according the response of your webhook.

I'm going to explain my use case:
I've written a custom webhook to implement a callback action for our messaging app where I only need set the alert status and send a message and no re-process alert. It is very similar to telegram webhook.
My custom webhook response is `jsonify` object and not an `Alert` object, so I expect it was the response and not raise an error.

This PR tryies to solve this problem checking if the incomming Alert is an `Alert` object to continue processing or return your custom response.

### DIFF
--- a/alerta/webhooks/custom.py
+++ b/alerta/webhooks/custom.py
@@ -5,6 +5,7 @@ from flask_cors import cross_origin
 from alerta.app import custom_webhooks
 from alerta.auth.decorators import permission
 from alerta.exceptions import ApiError, RejectException
+from alerta.models.alert import Alert
 from alerta.models.enums import Scope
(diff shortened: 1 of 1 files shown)


satterly:
Nice one. I hadn't considered that custom webhooks might be used for more than just ingesting alerts. The only problem I see is that the `write_audit_trail.send()` method is never called if it's not an alert.

satterly:
Thanks. I'll merge this as-is but I might need to refactor it slightly because I'll have to update the type annotations for webhook base class.

https://github.com/alerta/alerta/blob/145fe22de8ea9967281a232222fe03c03c0b1d8d/alerta/webhooks/__init__.py#L33

PR #759, opened 2018-11-21
Add query filter for users and perms

satterly:
"Flat" RBAC requires ...

  * user list to be filtered by role; and
  * perms list to be filtered by scope

.. which will enable admins to answer...

  * which users have been assigned a specific role; and
  * which roles have been assigned a specific permission (scope)

**Example - Which roles have the "write" scope?**
```
http://api.local.alerta.io:8080/perms?scopes=write
```
```json
{
  "permissions": [
    {
      "href": "http://api.local.alerta.io:8080/perm/18516f71-0309-44b2-bc53-166287e2c123",
      "id": "18516f71-0309-44b2-bc53-166287e2c123",
      "match": "read-write",
      "scopes": [
        "write"
      ]
    },
    {
      "href": "http://api.local.alerta.io:8080/perm/815eac5f-c6c8-4db5-b36c-26d76160b476",
      "id": "815eac5f-c6c8-4db5-b36c-26d76160b476",
      "match": "user",
      "scopes": [
        "read",
        "write"
      ]
    }
  ],
  "status": "ok",
  "total": 2
}
```

### DIFF
--- a/alerta/database/backends/postgres/utils.py
+++ b/alerta/database/backends/postgres/utils.py
@@ -104,7 +104,7 @@ class QueryBuilderImpl(QueryBuilder):
             if field in EXCLUDE_QUERY:
                 continue
             value = params.getlist(field)
-            if field in ['service', 'tags']:
+            if field in ['service', 'tags', 'roles', 'scopes']:
                 query.append('AND {0} && %({0})s'.format(field))
                 qvars[field] = value
             elif field.startswith('attributes.'):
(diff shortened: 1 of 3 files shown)


satterly:
There's an issue here when querying `/users` with `?roles=user` because "user" is the default role users aren't being found if no role defined in database and admin users are included if their default role is user but they are included in the `ADMIN_USERS` setting.

satterly:
[RBAC fixes](https://trello.com/c/B238cCZe/512-rbac-fixes)

PR #807, opened 2019-01-11
Prometheus Webhook: Add customer from labels

abhishekjiitr:
Adds **customer** information in an alert, extracted from **customer** label in Prometheus Alertmanager webhook. If no **customer** label is there customer is assigned _None_.


satterly:
Thanks. 👍

PR #809, opened 2019-01-12
Fix action-driven state transitions

satterly:


satterly:
[blackout status rethink](https://trello.com/c/AyIOXbzy/523-blackout-status-rethink)

PR #813, opened 2019-01-14
Lock version of pyparsing to 2.3.0

satterly:
Something has changed with PyParsing v2.3.1 which is causing builds to fail.

https://github.com/pyparsing/pyparsing/compare/pyparsing_2.3.0..pyparsing_2.3.1


satterly:
This change broke Alerta ...

https://github.com/pyparsing/pyparsing/commit/45b78401e8c224619e1b18b9cf43fc02b196676e

PR #827, opened 2019-01-23
Fix unconfigured log file handler

satterly:
Fixes alerta/docker-alerta#105


satterly:
I've tested the docker container by setting `LOG_FILE='/dev/null'` and it starts up ok. 👍

PR #836, opened 2019-02-04
Rancher support

sauber:
Accept alerts submitted from Rancher's alertmanager

### DIFF
--- a/alerta/webhooks/prometheus.py
+++ b/alerta/webhooks/prometheus.py
@@ -60,7 +60,7 @@ def parse_prometheus(alert: JSON, external_url: str) -> Alert:
 
     # labels
     resource = labels.pop('exported_instance', None) or labels.pop('instance', 'n/a')
-    event = labels.pop('event', None) or labels.pop('alertname')
+    event = labels.pop('event', None) or labels.pop('alertname', None) or labels.pop('alert_name')
     environment = labels.pop('environment', 'Production')
     customer = labels.pop('customer', None)
     correlate = labels.pop('correlate').split(',') if 'correlate' in labels else None


satterly:
Instead of modifying the Prometheus webhook to accept Rancher alerts I would suggest creating a new webhook called "rancher" and submit it to the alerta-contrib repo. Otherwise there are going to be unnecessary ongoing dependencies between the two projects going forward. See https://github.com/alerta/alerta-contrib/tree/master/webhooks for lots of examples.

sauber:
@satterly, thanks for reviewing and suggesting. I think it's a better long term approach to add to alerta-contrib. I have trouble enabling contrib webhooks. Can you suggest how to setup a developerment environment, and how to include contrib webhooks when using docker image https://github.com/alerta/docker-alerta ?

satterly:
I do all my development in a Python virtual environment. You can find out more about "[venv](https://docs.python.org/3/library/venv.html)" here.

Create a virtualenv then "pip install alerta-server" and git clone the "alerta-contrib" repo, cd into your rancher directory and create the necessary files and do "pip install -e ." so that "alertad" can find it.

To run a development environment ...
```
$ pip install alerta-server
$ git clone ...
$ cd alerta-contrib/webhooks/rancher
(create setup.py, alerta_rancher.py, test_rancher.py, README.md)
$ pip install -e .
$ alertad run
```

To run the tests (if using postgres) ...
```
$ cd alerta-contrib
$ pip install -r requirements-dev.txt
$ createdb test5
$ ALERTA_SVR_CONF_FILE= DATABASE_URL=postgres:///test5 pytest -v webhooks/rancher
```

To install a custom webhook into a docker container use the pattern which adds an additional RUN command layer to the default Alerta docker container. See for example [python-ldap docker install](https://github.com/alerta/docker-alerta/blob/master/examples/ldap/Dockerfile).

Let me know if you need more info.

PR #841, opened 2019-02-14
Fix the handling of % in webhook bodies

craig-miskell-fluxfederation:
Without this, % in the webhook bodies causes logging failures when the webhook body contains a %X that gets interpreted as a formatter by the python logging package (msg = msg % self.args).

In more detail: before_request in webhooks/__init__.py creates the message and passes it to the logger, but in the logger itself it also formats again (using %) *if* there are any args.  Interestingly, the args are  query params, so the previous tests didn't trigger this because there were none.  We triggered this in production because we were using api-key as a query param.  This is why the tests need to also have api-key changes.  

Also of note: the updated tests don't strictly fail without the fix in webhooks/__init__.py, they just write to stdout (the logging framework eats the formatting exception and spits out the entire failure to stdout); I don't know of any trivial way to make this properly fail in a test compatible fashion.  Also, I was unable to get it to fail/log poorly on anything other than the cloudwatch test (specifically, the first test).  I suspect something nose-related rather than the code, but I'm really not sure.  I am very happy to be enlightened if I'm missing something.

### DIFF
--- a/alerta/webhooks/__init__.py
+++ b/alerta/webhooks/__init__.py
@@ -21,7 +21,7 @@ def before_request():
         request.method,
         request.url,
         request.headers,
-        request.get_data().decode('utf-8')
+        request.get_data().decode('utf-8').replace("%","%%")
     ))
 
 
--- a/tests/test_webhooks.py
+++ b/tests/test_webhooks.py
@@ -6,6 +6,7 @@ from uuid import uuid4
 
 from alerta.app import create_app, custom_webhooks, db
 from alerta.models.alert import Alert
+from alerta.models.key import ApiKey
 from alerta.webhooks import WebhookBase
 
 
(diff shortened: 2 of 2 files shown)


satterly:
Thanks for this comprehensive explanation and suggested fix. Though instead of 'percent-encoding' the request payload just for webhooks a more general solution would be to disable the default Flask log handler. The default handler uses percent encoding for log format which is the root cause of this issue.

And it looks like I **was** going to do this as the code to remove the default handler exists but has been commented out (unfortunately I can't remember why so I think there could be side-effects).

https://github.com/alerta/alerta/blob/78e6e0654a1c67217c4f7396c7af6710d22a961d/alerta/utils/logging.py#L20-L21

Extending the test case to include the API key is a good idea. I couldn't get the test to fail either.

Do you want to modify this PR to uncomment line 21 above, or shall I do it? I'm happy either way.

satterly:
The actual problem turns out to have been the use of "args" as a record attribute when storing the request arguments in the logging filter. Just had to rename that to "reqargs" instead of "args".

craig-miskell-fluxfederation:
Oh nice, that's much simpler :)

PR #842, opened 2019-02-15
logger attribute

sauber:
Changing app.log to app.logger to avoid internal server error when loading broken webhook.


satterly:
Good catch. Thanks!

PR #843, opened 2019-02-15
Return code for non-alert reponses

sauber:
custom() must return http code as second paramenter.

### DIFF
--- a/alerta/webhooks/custom.py
+++ b/alerta/webhooks/custom.py
@@ -52,4 +52,4 @@ def custom(webhook):
         write_audit_trail.send(current_app._get_current_object(), event='webhook-received', message=text, user=g.user,
                                customers=g.customers, scopes=g.scopes, resource_id=None, type='user-defined',
                                request=request)
-        return jsonify(**response)
+        return jsonify(**response), 201


satterly:
Do you get a '200 OK' response at the moment? Because HTTP status code '201 Created' is only meant to be used if you create a new resource.

sauber:
> Do you get a '200 OK' response at the moment? Because HTTP status code '201 Created' is only meant to be used if you create a new resource.

When omittet, I get:

{
  "code": 404,
  "errors": null,
  "message": "Custom webhook 'rancher' not found.",
  "status": "error"
}

satterly:
Use `ApiError` exception to return non-2xx responses.

sauber:
> Use `ApiError` exception to return non-2xx responses.

OK, that will be fine. Here is my current structure:

    def incoming(self, query_string, payload):
        alerts = []
        if 'alerts' in payload:
            for alert in payload['alerts']:
                try:
                    self.parse_rancher(alert, external_url)
                except ValueError as e:
                    raise ApiError(str(e), 400)
                alerts.append(alert)
        else:
            raise ApiError('no alerts in Rancher notification payload', 400)

        if len(alerts) == 1:
            return {'status': 'ok', 'id': alerts[0].id, 'alert': alerts[0].serialize}
        else:
            return {'status': 'ok', 'ids': [alert.id for alert in alerts]}

So I think the parent can assume return code 201.

satterly:
Looks good except the last few lines should be like the prometheus webhook ...
```

    if len(alerts) == 1:
        return jsonify(status='ok', id=alerts[0].id, alert=alerts[0].serialize), 201
    else:
        return jsonify(status='ok', ids=[alert.id for alert in alerts]), 201
```
You need to use `jsonify()` otherwise you're not returning a Flask response.

sauber:
> Looks good except the last few lines should be like the prometheus webhook ...
> 
> ```
> 
>     if len(alerts) == 1:
>         return jsonify(status='ok', id=alerts[0].id, alert=alerts[0].serialize), 201
>     else:
>         return jsonify(status='ok', ids=[alert.id for alert in alerts]), 201
> ```
> You need to use `jsonify()` otherwise you're not returning a Flask response.

In that case the response is jsonify()'ed twice: inside the custom webhook, and again inside custom.py. And python objects with this greeting: 

return jsonify(**response)\nTypeError: jsonify() argument after ** must be a mapping, not tuple\n

sauber:
The non-customer webhooks can get way with jsonify the reponse, must even, because they are not wrapped by custom.py .

Honestly, it's a mess that first class webhooks, and custom webhooks are have different interfaces.

satterly:
Sorry, I got confused because I'm jumping between three different PR/Issues related to this. You were right the first time. If you don't return an `Alert` then it the custom webhook interface will convert whatever you return (using `jsonify()`) into a Flask Response.

I'm closing this PR because it's not relevant and just confusing.

sauber:
https://github.com/alerta/alerta/pull/844 seems a much better solution. Thanks for addressing.

PR #845, opened 2019-02-16
Fix logging request filter "args" name clash

### LINKED ISSUE #481: How to delete a tag from alert specially when alert has history
I have alert1 coming into alerta with a tag of "key1=val1". 
alert1 is then cleared and alerta makes it closed. So far all good.

Now alert1 is coming ito alerta with a tag of "key1=val2". In my logs in plugins I can verify that it is the tag i'm sending it with. However in the UI it shows both "key1=val1" and "key1=val2" as tags. Only way around it is to remove the tag from alerta by deleting the alert which is discouraged.

Please let me know how can I make remove that tag from alert history so that it doesn't affect my plugin logic based on those tags.

satterly:
Fixes #841

### DIFF
--- a/alerta/utils/logging.py
+++ b/alerta/utils/logging.py
@@ -45,6 +45,7 @@ class Logger:
 
             dictConfig({
                 'version': 1,
+                'disable_existing_loggers': False,
                 'formatters': {
                     'default': {
                         'format': logging.BASIC_FORMAT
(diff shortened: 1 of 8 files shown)


alerta/utils/logging.py
```diff
@@ -109,7 +110,7 @@ def filter(self, record):
             record.endpoint = request.endpoint
             record.method = request.method
             record.url = request.url
-            record.args = request.args
+            record.reqargs = request.args
```
satterly: This is the fix.

PR #852, opened 2019-02-22
mongo base warning

sauber:
Address warning:

alerta/alerta/database/backends/mongodb/base.py:32: DeprecationWarning: get_default_database is deprecated. Use get_database instead.
    return self.client.get_default_database()

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -29,7 +29,7 @@ class Backend(Database):
         if self.dbname:
             return self.client[self.dbname]
         else:
-            return self.client.get_default_database()
+            return self.client.get_database()
 
     @staticmethod
     def _create_indexes(db):


satterly:
Thanks. I've been meaning to fix this but needed to make sure it didn't break old behaviour. Need to double check first.

satterly:
Looks like the behaviour of `get_database()` was changed to replicate the behaviour of 
 `get_default_database()` from version 3.5. So perhaps the minimum version number for pymongo should be changed to `>=3.5` ... 
https://github.com/alerta/alerta/blob/24e7eb2fe5458a4b5dc5c2dbac6f3be9bdb53876/setup.py#L44

>Changed in version 3.5: The name parameter is now optional, defaulting to the database named in the MongoDB connection URI.

http://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient.get_database

satterly:
So pymongo 3.5 was released 18 months ago. There's a risk that some OS distributions might not have at least this version packaged yet. It's worth investigating Debian/Ubuntu and Centos to see what versions they ship with. Note that `get_default_database()` has been deprecated but no date/version for removal has been given.

satterly:
The minimum version of Python 3 for Alerta is 3.5. The first version of pymongo to support version 3.5 or later was pymongo version 3.6.0. So I think this means there's no way anyone could be running a version of Alerta 6.x that does not have pymongo >=3.5 which means we can merge this without fear.

PR #853, opened 2019-02-24
Oembed fix

kak-tus:
Oembed support seems to be broken.
May be in case of running alerta in docker+postgresql. May be entirelly.

As example: /api/oembed endpoint accessible as /api/oembed, but not as /api/oembed.json.
Also: qb.from_params forms incorrect query to postgresql.

I am not a python programmer and don't know flask framework, so I fixed this not ideally.

After this fixes I've got oembed work in grafana
image

### DIFF
--- a/alerta/views/oembed.py
+++ b/alerta/views/oembed.py
@@ -28,9 +28,7 @@ def oembed(format):
 
     try:
         url = request.args['url']
-        width = int(request.args['maxwidth'])
-        height = int(request.args['maxheight'])
-        title = request.args.get('title', 'Alerts')
+        title = request.args['title']
     except Exception as e:
         return jsonify(status='error', message=str(e)), 400
 
@@ -59,15 +57,13 @@ def oembed(format):
         html = render_template(
             'oembed/counts.html',
             title=title,
-            width=width,
-            height=height,
             max=max,
             counts=severity_count
         )
         headers =	{
           "Access-Control-Allow-Origin": "*"
         }
-        return jsonify(version='1.0', type='rich', width=width, height=height, title=title, provider_name='Alerta', provider_url=request.url_root, html=html), 200, headers
+        return jsonify(version='1.0', type='rich', title=title, provider_name='Alerta', provider_url=request.url_root, html=html), 200, headers
 
     elif o.path.endswith('/alerts/top10/count'):
         # TODO: support top10 oembed widget
--- a/alerta/views/oembed.py
+++ b/alerta/views/oembed.py
@@ -9,8 +9,11 @@ from alerta.auth.decorators import permission
 from alerta.models.enums import Scope
 from alerta.utils.response import jsonp
 
+from collections import namedtuple
+
 from . import api
 
+Query = namedtuple('Query', ['where', 'vars', 'sort', 'group'])
 
 @api.route('/oembed', defaults={'format': 'json'}, methods=['OPTIONS', 'GET'])
 @api.route('/oembed.<format>', methods=['OPTIONS', 'GET'])
@@ -18,14 +21,6 @@ from . import api
 @permission(Scope.read_oembed)
 @jsonp
 def oembed(format):
-
-    if format != 'json':
-        return jsonify(status='error', message='unsupported format: %s' % format), 400
-
-    if 'url' not in request.args or 'maxwidth' not in request.args \
-            or 'maxheight' not in request.args:
-        return jsonify(status='error', message='missing default parameters: url, maxwidth, maxheight'), 400
-
     try:
         url = request.args['url']
         title = request.args['title']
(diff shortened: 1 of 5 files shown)


examples/oembed.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <link rel="stylesheet" href="//netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css" />
+  <link rel="stylesheet" href="http://netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css" />
```
vykulakov: Why have you changed `//` to `http://` here and in other places of code? `//` looks like a more general approach that will work in a wider range of environments. For instance, I'm trying to access the `oembed.html` page bu HTTPs in my experiments.

examples/oembed.html
```diff
@@ -13,22 +13,27 @@
 
 <div class="mobile-alerts">Loading...</div>
 
-<script src="//ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
-<script src="//netdna.bootstrapcdn.com/bootstrap/3.3.2/js/bootstrap.min.js"></script>
-<script src="http://api.alerta.io/embed.js"></script>
+<script src="http://ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
+<script src="http://netdna.bootstrapcdn.com/bootstrap/3.3.2/js/bootstrap.min.js"></script>
+<script src="http://localhost:8080/api/embed.js"></script>
```
vykulakov: Using the `/api` path prefix here is not correct because it will work only in your environment. Please, use more general URL like this: `//localhost:8080/embed.js`

examples/oembed.html
```diff
@@ -13,22 +13,27 @@
...
 <script>
-$(function() {
+$(window).on('load', function () {
 
   // $.alerta.defaults.endpoint = 'http://localhost:8080';
   // $.alerta.defaults.key = 'demo-key';
   // or
   $.alerta.defaults = {
-      endpoint: 'http://api.alerta.io', // oEmbed endpoint becomes http://api.alerta.io/oembed.json
+      endpoint: 'http://localhost:8080/api', // oEmbed endpoint becomes http://localhost:8080/api/oembed
```
vykulakov: See the comment about `/api` above.

examples/oembed.html
kak-tus: To get ability to user open this page locally and test with local instance of alerta.
Without this fix, user opens page with "file:///" protocol and external scripts tries to open with "file:///" protocol too.
If this page is not test page, but it is an example page - I will revert this fixes.

examples/oembed.html
kak-tus: Fixed.

examples/oembed.html
kak-tus: Fixed.

examples/oembed.html
satterly: Yes, please revert this change.

examples/oembed.html
```diff
@@ -1,29 +1,44 @@
...
   // $.alerta.defaults.key = 'demo-key';
   // or
   $.alerta.defaults = {
-      endpoint: 'http://api.alerta.io', // oEmbed endpoint becomes http://api.alerta.io/oembed.json
+      endpoint: 'http://localhost:8080', // oEmbed endpoint becomes http://localhost:8080/oembed
       key: 'demo-key'
   };
 
-  $('.mobile-alerts').alerta('http://api.alerta.io/alerts/count?service=Mobile&status=open', {title:'Mobile Service'});
+  function renew() {
```
satterly: Can you call this function `refresh()` instead of `renew()`? Thanks.

alerta/views/oembed.py
```diff
@@ -9,28 +9,21 @@
...
-    if 'url' not in request.args or 'maxwidth' not in request.args \
-            or 'maxheight' not in request.args:
-        return jsonify(status='error', message='missing default parameters: url, maxwidth, maxheight'), 400
-
     try:
         url = request.args['url']
-        width = int(request.args['maxwidth'])
-        height = int(request.args['maxheight'])
-        title = request.args.get('title', 'Alerts')
+        title = request.args['title']
```
satterly: Why did you remove the default title? It will throw an exception here now if someone does not provide `title` as a URL param.

satterly:
Thanks for the PR. It looks good except for a few minor changes. I'll be happy to merge this once you've addressed the concerns I've raised above. 👍

vykulakov:
Looks like I've forgotten to submit my review earlier. My mistake, sorry.

PR #854, opened 2019-02-26
get NORMAL_SEVERITY_LEVEL from DEFAULT_NORMAL_SEVERITY instead of har…

wimfabri:
get NORMAL_SEVERITY_LEVEL from DEFAULT_NORMAL_SEVERITY instead of hardcoded to 5


satterly:
Make sense. Good work. Thanks. 👍

PR #859, opened 2019-02-28
fix for bulk-acknowledged alert re-opening

wimfabri:


satterly:
Thanks. Looking at this made me realise that the web UI is using the wrong resource endpoint for bulk actions. ie. the old "status" endpoint instead of "action".

I've made the change to fix that however I think it's still important to understand why not setting "value" causes the error to surface.

satterly:
Having looked at this in more detail your fix is still required to avoid "duplicate alert with value change" updates to alert history when the value hasn't actually changed at all. I'll merge this and then fix the test in another PR. Thanks a lot. 👍

Issue #906, opened 2019-04-01
AMQP plugin breaks adding alerts


ajgon:
**Issue Summary**
After installing AMQP plugin, I am unable to add any event to alerta backend.

**Environment**
- OS: Linux
- API version: 6.8.3
- Deployment: Docker

- Database: Postgres

- Env variables
```ini
ADMIN_KEY=...
ADMIN_PASSWORD=...
ADMIN_USERS=alerta@test.host
ALERTA_CONF_FILE=/app/alerta.conf
ALERTA_SVR_CONF_FILE=/app/alertad.conf
ALERTA_WEB_CONF_FILE=/web/config.json
AMQP_TOPIC=alerta.notify
AUTH_REQUIRED=True
BASE_URL=/api
DATABASE_URL=postgres://...
GPG_KEY=[...]
HEARTBEAT_SEVERITY=major
INSTALL_PLUGINS=normalise,enhance,amqp
LANG=C.UTF-8
MAIL_FROM=alerta@test.host
PATH=/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/venv/bin
PLUGINS=reject,blackout,normalise,enhance,amqp
PYTHON_PIP_VERSION=19.0.3
PYTHONUNBUFFERED=1
PYTHON_VERSION=3.6.8
REDIS_URL=redis://redis:6379/
```

- Server config:
```json
{
    "actions": [],
    "alarm_model": {
        "colors": {
            "highlight": "skyblue ",
            "severity": {
                "cleared": "#00CC00",
                "critical": "red",
                "debug": "#9D006D",
                "indeterminate": "lightblue",
                "informational": "#00CC00",
                "major": "orange",
                "minor": "yellow",
                "normal": "#00CC00",
                "ok": "#00CC00",
                "security": "blue",
                "trace": "#7554BF",
                "unknown": "silver",
                "warning": "dodgerblue"
            },
            "text": "black"
        },
        "name": "Alerta 6.8.3",
        "severity": {
            "cleared": 7,
            "critical": 1,
            "debug": 8,
            "indeterminate": 5,
            "informational": 6,
            "major": 2,
            "minor": 3,
            "normal": 7,
            "ok": 7,
            "security": 0,
            "trace": 9,
            "unknown": 10,
            "warning": 4
        },
        "status": {
            "ack": "C",
            "assign": "B",
            "blackout": "E",
            "closed": "F",
            "expired": "G",
[...]

PR #924, opened 2019-04-08
 Issue with alert serialisation 

ashwinbasavaraja:
Currently when the new alerts getting created in 'updateTime' will be None. So it create issue as described -https://github.com/alerta/alerta/issues/906


satterly:
Thanks for this but the easier fix is just to skip datetime attributes that are not set yet. See https://github.com/alerta/alerta/blob/master/alerta/models/alert.py#L157

The problem was compounded by the fact that the "update_time" column type for Postgres was wrong. I've also added a test that should catch any issues like this in future. See #925

PR #940, opened 2019-04-16
Fix status change hook alert context

### LINKED ISSUE #939: Add attributes of an origin alert to a correlated alert

I'm trying to modify the standard alerta-prometheus plugin to remove a silence period in AlertManager when Alerta receives a corresponding normal (or cleared) alert. Now it is not possible because the normal alert in the `status_change` handler (https://github.com/alerta/alerta-contrib/blob/master/plugins/prometheus/alerta_prometheus.py#L37) does not contain attributes from an original alert. The attributes are added to this alert right after `status_change_hook` is completed (https://github.com/alerta/alerta/blob/master/alerta/models/alert.py#L350).

Is it possible to execute the status changed hook after getting the correlated alert with filled attributes from DB?

(don't know)

**Additional context**

slava @vykulakov Apr 15 11:29
@satterly I've tried to describe my problem here: alerta/python-alerta-client#182
The problem is that a corresponding normal or cleared alert which closes an original alert does not contain attributes from the original alert.
So I cannot get silenceId in the status_change handler.

slava @vykulakov Apr 15 11:38
There is the code:
        if new_status != status:
            status_change_hook.send(self, status=new_status, text=self.text)
            self.update_time = now

        self.status = new_status
        return Alert.from_db(db.correlate_alert(self, history))
If I understood correctly, the last line loads the original and updated alert from DB with attributes and it will be great if we invoke status_change_hook.send() after this line with the loaded and updated alert.

slava @vykulakov Apr 15 12:04
Hm... But in this case, we won't store silenceId to attributes without additional updating.

slava @vykulakov Apr 15 12:17
Oh, I see. I can use the post_receive handler to remove the silence period on receiving the normal alert... But it is not so convenient. And anyway, I must invoke additional updating to remove silenceId from attributes in DB.

Nick Satterly @satterly Apr 15 13:53
I don't understand why you are updating silences on closed. Surely that is managed by Alertmanager before it is received by Alerta. If anything the problem exists in reverse. Where Alertmanager fails to update subscribers when a silenced alert clears.
alerta/alerta#278
But.you know about this already so I'm confused

slava @vykulakov Apr 15 14:00
Yes, I understand why you are confused now. I must give you some details again. I have two AlertManager instances now. One to send notification via email and slack and one to send alerts to Alerta. So with this configuration Alerta can silence alerts and nobody will receive them but Alerta will continue receiving alerts from the second instance.

slava @vykulakov Apr 15 14:09
This configuration solves the problem, described in alerta/alerta#278. It is not so beautiful but it works and I have actual data in Alerta without any timeouts, housekeeping and so on.
But the problem described above should be actual not only for Prometheus and AlertManager because handlers don't have all the information about alerts. It is quite general.

satterly:
For duplicate or correlated alerts, run the status change hook in the context of the duplicate or correlated alert from the database, not the alert that was sent. This means that any attributes or tags (or other alert information) associated with the existing alert is accessible to the status change plugin.

However, it means any attributes or tags in the sent alert are not accessible. This could be a problem, though all the tests pass so it's not a use-case I am familiar with. This is not to say anyone isn't using status change plugin in this way already so this could be creating more problems than it solves.

Fixes #939

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -121,7 +121,8 @@ class Backend(Database):
             'severity': alert.severity,
             'customer': alert.customer
         }
-        return bool(self.get_db().alerts.find_one(query))
+        r = self.get_db().alerts.find_one(query, projection={'_id': 1})
+        return r['_id'] if r else False
 
     def is_correlated(self, alert):
         query = {
@@ -138,7 +139,8 @@ class Backend(Database):
                 }],
             'customer': alert.customer
         }
-        return bool(self.get_db().alerts.find_one(query))
+        r = self.get_db().alerts.find_one(query, projection={'_id': 1})
+        return r['_id'] if r else False
 
     def is_flapping(self, alert, window=1800, count=2):
         """
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -121,8 +121,7 @@ class Backend(Database):
             'severity': alert.severity,
             'customer': alert.customer
         }
-        r = self.get_db().alerts.find_one(query, projection={'_id': 1})
-        return r['_id'] if r else False
+        return self.get_db().alerts.find_one(query)
 
     def is_correlated(self, alert):
         query = {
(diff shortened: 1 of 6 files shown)


satterly:
Made some changes to reduce the number of DB calls. Now the only additional DB calls are to update alert tags and attributes and only if there is a status change.

vykulakov:
Looks fine. And as I can see you've made some additional changes to process `new_status` and `text` from a handler. Right?

satterly:
That's right. Which is why there are some changes to the test cases -- which I need to do a final review of to make sure it makes sense that they have changed / been updated.

satterly:
Different results when running test cases seem correct so the test cases have been updated to reflect changed behaviour. Hopefully this bug fix won't break plugins that rely on the "broken" way of working.

PR #944, opened 2019-04-20
Push app config to plugin methods for per-alert config

### LINKED ISSUE #746: Customer based routing plugins' configuration
Currently all the plugins in Alerta run with same configuration. Due to which different customers having same routing plugin but different configuration can't be boarded. An example might be:

1. If `customer=='A'` route to slack with API key `123456` on slack channel `#alerts`
2. If `customer=='B'` route to slack with API key `789012` on slack channel `#infra-alerts`

(minor discussion on the same can be found in Gitter [here](https://gitter.im/alerta/chat?at=5be940ad4e7ca14520920647))

The routing should be based on alerts (which already happen in Alerta) but at the same time, plugins can dynamically change their configuration. For ex: if the `alert.owner` is `customerA` then plugin should be running with tokens and other configurations defined for `customerA` and so on. (default config for configs not provided/overridden)

I am willing to know views/alternative solution design, but my initial thoughts are as follows which I believe to be backward compactible (taking `opsgenie` plugin as example):

1. `settings.py`
```
OPSGENIE_SERVICE_KEY = 'abcdefghijkl'
SERVICE_KEY_MATCHERS = []
CUSTOMER_CONFIG_OVERRIDES = {
    'customerA': {
        'opsgenie': {
            'OPSGENIE_SERVICE_KEY': '1234567890'
        }
    },
}
```

2. `alerta_opsgenie.py`

Currently, variables are defined as module globals and used in class directly (`OPSGENIE_SERVICE_KEY`, etc). This can be updated in plugins to `self.get_config('OPSGENIE_SERVICE_KEY', alert.owner)`, while `PluginBase` can have `get_config` function which return default or overridden config based on `alert.owner` value.

One alternative can be initializing plugins multiple times i.e. one for all customers in `utils/plugin.py` and directly use that instance based on `alert.owner` or default if `alert.owner is None`.

satterly:
Fixes #746

### DIFF
--- a/alerta/plugins/blackout.py
+++ b/alerta/plugins/blackout.py
@@ -2,11 +2,9 @@ import logging
 import os
 
 from alerta.exceptions import BlackoutPeriod
-from alerta.plugins import PluginBase, app
+from alerta.plugins import PluginBase
 
 LOG = logging.getLogger('alerta.plugins')
-
-NOTIFICATION_BLACKOUT = os.environ.get('NOTIFICATION_BLACKOUT') or app.config.get('NOTIFICATION_BLACKOUT', True)
 
 
 class BlackoutHandler(PluginBase):
@@ -19,12 +17,16 @@ class BlackoutHandler(PluginBase):
     any plugins for further notification.
     """
 
-    def pre_receive(self, alert):
+    def pre_receive(self, alert, **kwargs):
 
-        if app.config['ALARM_MODEL'] == 'ALERTA':
+        config = kwargs['config']
+
+        if config['ALARM_MODEL'] == 'ALERTA':
             status = 'blackout'
         else:
             status = 'OOSRV'  # ISA_18_2
+
+        NOTIFICATION_BLACKOUT = os.environ.get('NOTIFICATION_BLACKOUT') or config.get('NOTIFICATION_BLACKOUT', True)
 
         if alert.is_blackout():
             if NOTIFICATION_BLACKOUT:
@@ -35,10 +37,10 @@ class BlackoutHandler(PluginBase):
                 raise BlackoutPeriod('Suppressed alert during blackout period')
         return alert
 
-    def post_receive(self, alert):
+    def post_receive(self, alert, **kwargs):
         return
 
-    def status_change(self, alert, status, text):
+    def status_change(self, alert, status, text, **kwargs):
         return
 
     def take_action(self, alert, action, text, **kwargs):
--- a/alerta/plugins/blackout.py
+++ b/alerta/plugins/blackout.py
@@ -1,5 +1,4 @@
 import logging
-import os
 
 from alerta.exceptions import BlackoutPeriod
 from alerta.plugins import PluginBase
@@ -18,15 +17,14 @@ class BlackoutHandler(PluginBase):
     """
 
     def pre_receive(self, alert, **kwargs):
+        self.kwargs = kwargs
 
-        config = kwargs['config']
+        NOTIFICATION_BLACKOUT = self.get_config('NOTIFICATION_BLACKOUT', default=True, type=bool)
 
-        if config['ALARM_MODEL'] == 'ALERTA':
+        if self.get_config('ALARM_MODEL') == 'ALERTA':
             status = 'blackout'
         else:
             status = 'OOSRV'  # ISA_18_2
-
-        NOTIFICATION_BLACKOUT = os.environ.get('NOTIFICATION_BLACKOUT') or config.get('NOTIFICATION_BLACKOUT', True)
 
         if alert.is_blackout():
             if NOTIFICATION_BLACKOUT:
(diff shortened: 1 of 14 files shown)


satterly:
[customer based plugin config](https://trello.com/c/0YqTLNhy/531-customer-based-plugin-config)

alerta/plugins/blackout.py
```diff
@@ -19,13 +17,17 @@ class BlackoutHandler(PluginBase):
...
+
+        config = kwargs['config']
 
-        if app.config['ALARM_MODEL'] == 'ALERTA':
+        if config['ALARM_MODEL'] == 'ALERTA':
             status = 'blackout'
         else:
             status = 'OOSRV'  # ISA_18_2
 
+        NOTIFICATION_BLACKOUT = os.environ.get('NOTIFICATION_BLACKOUT') or config.get('NOTIFICATION_BLACKOUT', True)
```
skbly7: Minor:
I was wondering if we can create a function like `get_config` added in `PluginBase` class which can extract configuration from `kwargs['config']`. 
This will ease refactoring of all the existing Alerta plugins. So, adding below code:

```
config = kwargs['config'] 
...
NOTIFICATION_BLACKOUT = os.environ.get('NOTIFICATION_BLACKOUT') or config.get('NOTIFICATION_BLACKOUT', True)
```

will just become

```
NOTIFICATION_BLACKOUT = self.get_config('NOTIFICATION_BLACKOUT')
```

satterly: Yes, I tried exactly this but couldn't get it work. If you've got any ideas let me know. 😢

skbly7: Hmm I see. We are transferring config as kwargs, so don't have much freedom, what about:

```
    def get_config(self, key, default, **kwargs):
        config = kwargs['config'] 
        return config.get(key) or os.environ.get(key) or default
```

And usage as:

```
    NOTIFICATION_BLACKOUT = self.get_config('NOTIFICATION_BLACKOUT', True, kwargs)
```

Also, I just realised the ordering is wrong in original commit. i.e. kwargs config should have higher preference, followed by env value and then default. (earlier was env > kwargs)

skbly7: Thanks a lot, the `get_config` method looks awesome, even better with `type` parameter. :D

satterly: Good suggestion. The helper method I've come up with is ...
```
def get_config(self, key, default=None, type=None)
```
And usage is ...
```
self.kwargs = kwargs

NOTIFICATION_BLACKOUT = self.get_config('NOTIFICATION_BLACKOUT', default=True, type=bool)
```
Note: you must initialise `self.kwargs` yourself if you want to use `self.get_config()`. It will raise an exception if you try to use the helper without initialising first.

If you supply a `type` it will convert any environment variable into that type. Supported types are `bool`, `int`, `float`, `list`, JSON-encoded `dict` with a default of `str`.

satterly: Oh, and precedence order is environment variable first, then any defined config, then default. I'm not sure why you would want to reverse it. If you disagree can you please elaborate?

skbly7: Yes, I think defined config should be first because if someone is especially passing config in `routing.py`, he would especially want it be used & knows what he is doing. While the env variable/default for the fallback alerts. (at least how I am using right now)

Otherwise, the feature will not be useful/workable if whole Alerta have an environment value defined.

alerta/plugins/blackout.py
```diff
@@ -19,9 +16,12 @@ class BlackoutHandler(PluginBase):
     any plugins for further notification.
     """
 
-    def pre_receive(self, alert):
+    def pre_receive(self, alert, **kwargs):
+        self.kwargs = kwargs
```
skbly7: Just realised, what will happen in case of multiple alerts (race condition)? 
Given we use one instantiation of plugin for all the alerts, it may cause wrong routing.

satterly: One per thread, but yes, potentially a race condition. 😞

satterly: I've gone with your original suggestion...
```
def get_config(self, key, default=None, type=None, **kwargs)
```
Usage ...
```
NOTIFICATION_BLACKOUT = self.get_config('NOTIFICATION_BLACKOUT', default=True, type=bool, **kwargs)
```
It's not quite as elegant as I wanted but in Python "explicit is better than implicit" so that's that.

alerta/plugins/blackout.py
satterly: I think there's a slight misunderstanding in the way this works. The only way to get config is via `kwargs` or environment variables so there is no difference (from a config priority point of view) between someone creating a routing rule that creates specific config and someone who doesn't.

Standard practice is to use environment variables as highest precedence which override all other config values. Sensible defaults should be set in the plugin and bespoke config values can be set in `alertad.conf` or the "routing rules".

I think that would work as most people would expect, and if this config precedence doesn't work for you then I would suggest that you aren't using environment variables correctly.

Happy to continue this discussion on https://gitter.im/alerta/chat.

satterly:
@skbly7 Thanks for reviewing the PR. Much appreciated. 👍

Issue #951, opened 2019-04-24
change the basedn formatter to have flexibility in query ldap


tiagosouza488:
**Is your feature request related to a problem? Please describe.**
The ldap bind formater (file alerta/auth/basic_ldap.py line 31) uses an old python string format that does not allow flexibility when formatting the DN for ldap binds.
On large ldap bases users are organized into subcontainers that can vary from base to base. For example in my ldap base my user stays in:
tiago_nascimento, cn = t, cn = users, dc = sicredi, dc = com, dc = br
Being that inside the container "t" are all users that start with the letter "t" and so on. With the current string formatter it is not possible to flex the assembly of the search dn which prevents the use of the Alerta with large ldap databases.

**Describe the solution you'd like**
Change line 31 in file "alerta/auth/basic_ldap.py" from:
userdn = current_app.config['LDAP_DOMAINS'][domain] % username
To:
userdn = current_app.config['LDAP_DOMAINS'][domain].format(username)

This way we can do this:
uid={0},cn={0:.1},cn=users,dc=sicredi,dc=com,dc=br
An many other customizations we need to configure dn in the right way.

**Describe alternatives you've considered**
Another solution to this problema is to separate the search base and the search filter into two different fields as other apps usually do. Having the search base and the search filter is done a query to get the dn and then the login is done with the dn obtained to validate the password. That would be a bit more complex to implement, but is the best solution to work with ldap databases in any format.

Ps.: I make a pull a request with the simple solution.

satterly:
>Another solution to this problema is to separate the search base and the search filter into two different fields as other apps usually do. Having the search base and the search filter is done a query to get the dn and then the login is done with the dn obtained to validate the password. 

Do you have an example of other apps that work this way? I'd like to take a look at how they define their config. Thanks.

satterly:
Closing due to inactivity. If you believe this issue has been closed prematurely please provide more information to help progress the issue. If you do not know what is expected you can ask on [Gitter chat](https://gitter.im/alerta/chat).

satterly:
This is already possible. See #771 and alerta/alerta-docs#95.

PR #952, opened 2019-04-24
Change LDAP base domain string format

tiagosouza488:
This is a simple solution suggest in the issue #951.

I don't find docs in this repo for adjusts de docs, but with this change th setup of LDAP_DOMAINS change from "uid=%s,cn=users,dc=mycompany,dc=com" to "uid={0},cn=users,dc=mycompany,dc=com"

### DIFF
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -28,7 +28,7 @@ def login():
     if domain not in current_app.config['LDAP_DOMAINS']:
         raise ApiError('unauthorized domain', 403)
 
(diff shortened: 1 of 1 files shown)


satterly:
This is a breaking change and I can't merge this without understanding why it can't be backward-compatible. How could this be modified to ensure that old style '%' works while also support "format" style string?

satterly:
Closing due to inactivity. If you believe this pull request has been closed prematurely please provide more information to help progress the issue. If you do not know what is expected you can ask on [Gitter chat](https://gitter.im/alerta/chat).

PR #954, opened 2019-04-25
Option to close not delete expired alerts

### LINKED ISSUE #887: Allow housekeeping to close but not delete stale alerts
> I use Alerta to display alerts from Prometheus/AlertManager. It works fine, but there are some cases when AlertManager does not send OK event. For instance, when we create a silence period for alerts and then these alerts are closing or when there are inhibit rules and there are inhibited alerts which are closing in that state. So I usually have some stale alerts in Alerta which I should close manually.
I have tried to use the housekeeping feature but it just removes stale alerts. It is not so good because of alerts history - I still want to see the stale alerts (which are closed/deleted now) in the history.
Is there any way to close the stale alerts instead of deleting them during housekeeping?
@vykulakov

satterly:
Fixes #887


satterly:
Closing this in favour of #970

PR #995, opened 2019-06-06
Azure oauth url changed to 2.0 & added openid scopes

erichester76:
Had issue with issuer_url validation after changing url. It also appears we made similar changes for the preferred_username however I compressed it all down to email since I didnt see the scenarios for nickname and/or email being different than login. I obviously dont have the exposure you do. I added the openid scopes. the URL and scopes seemed to fix the problem for me. Im honestly not sure if both were required but the new URL seems more proper, and the lack of profile in the scope according to documentation was definitely an issue. I am not sure if this will cause problems with other oauth providers since I am only testing azure currently.

### DIFF
--- a/alerta/auth/oidc.py
+++ b/alerta/auth/oidc.py
@@ -18,7 +18,7 @@ from . import auth
 def get_oidc_configuration(app):
 
     OIDC_ISSUER_URL_BY_PROVIDER = {
-        'azure': 'https://sts.windows.net/{}/'.format(app.config['AZURE_TENANT']),
+        'azure': 'https://login.microsoftonline.com//{}/v2.0'.format(app.config['AZURE_TENANT']),
         'gitlab': app.config['GITLAB_URL'],
         'google': 'https://accounts.google.com',
         'keycloak': '{}/auth/realms/{}'.format(app.config['KEYCLOAK_URL'], app.config['KEYCLOAK_REALM'])
@@ -35,8 +35,8 @@ def get_oidc_configuration(app):
     except Exception as e:
         raise ApiError('Could not get OpenID configuration from well known URL: {}'.format(str(e)), 503)
 
-    if config['issuer'] != issuer_url:
-        raise ApiError('Issuer Claim does not match Issuer URL used to retrieve OpenID configuration', 503)
+    #if config['issuer'] != issuer_url:
+    #    raise ApiError('Issuer Claim does not match Issuer URL used to retrieve OpenID configuration: {}'.format(issuer_url), 503)
 
     if app.config['OIDC_VERIFY_TOKEN']:
         try:
@@ -69,6 +69,7 @@ def openid():
         'redirect_uri': request.json['redirectUri'],
         'client_id': request.json['clientId'],
         'client_secret': current_app.config['OAUTH2_CLIENT_SECRET'],
+        'scope' : 'openid profile email',
     }
     r = requests.post(token_endpoint, data)
     token = r.json()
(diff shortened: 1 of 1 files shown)


erichester76:
I also really don't seem to be getting the interaction between the api code and the new webui code which has conflicting config in it in alerta-webu:i/src/services/auth.ts as my changes seemed to make a difference yet that code had the right scopes and url.

alerta/auth/oidc.py
```diff
@@ -114,7 +115,7 @@ def openid():
         group_claim: userinfo.get(group_claim) or id_token.get(group_claim, []),
     }
 
-    login = username or nickname or email
+    login = email
```
satterly: This is required to support OpenID connect providers that don't supply any of `username`, `nickname` or `email`. The Alerta auth flow needs at least one but only relying on email will definitely fail for lots of people.

alerta/auth/oidc.py
```diff
@@ -69,6 +69,7 @@ def openid():
         'redirect_uri': request.json['redirectUri'],
         'client_id': request.json['clientId'],
         'client_secret': current_app.config['OAUTH2_CLIENT_SECRET'],
+        'scope' : 'openid profile email',
```
satterly: Where in the spec does it suggest that scopes should be sent? This shouldn't be required.

alerta/auth/oidc.py
```diff
@@ -103,7 +104,7 @@ def openid():
     name = userinfo.get('name') or id_token.get('name')
     username = userinfo.get('preferred_username') or id_token.get('preferred_username')
     nickname = userinfo.get('nickname') or id_token.get('nickname')
-    email = userinfo.get('email') or id_token.get('email')
+    email = userinfo.get('email') or id_token.get('email') or username
```
satterly: I think it's confusing to assign username to email if email is undefined.

alerta/auth/oidc.py
```diff
@@ -35,8 +35,8 @@ def get_oidc_configuration(app):
     except Exception as e:
         raise ApiError('Could not get OpenID configuration from well known URL: {}'.format(str(e)), 503)
 
-    if config['issuer'] != issuer_url:
-        raise ApiError('Issuer Claim does not match Issuer URL used to retrieve OpenID configuration', 503)
+    #if config['issuer'] != issuer_url:
```
satterly: We need to understand why this verification of the issuer fails rather than just comment it out.

alerta/auth/oidc.py
```diff
@@ -18,7 +18,7 @@
 def get_oidc_configuration(app):
 
     OIDC_ISSUER_URL_BY_PROVIDER = {
-        'azure': 'https://sts.windows.net/{}/'.format(app.config['AZURE_TENANT']),
+        'azure': 'https://login.microsoftonline.com//{}/v2.0'.format(app.config['AZURE_TENANT']),
```
satterly: There's an extra slash between `.com` and the tenant id which is preventing the issuer URL from matching.

alerta/auth/oidc.py
satterly: See comment above about extra slash. This should be uncommented.

erichester76:
Right but look above in the patch and you'll see it set email to username or email. Same end result. Not sure you need that change. Your last patch made a change too and we collided I think.

satterly:
I've got it to work using a "consumer" application and a tenant application so I'm pretty confident it would work for you too. It required the change to v2 api (as you did) but also a small change to the web UI authorization endpoint URL for azure.

Can you revert all your changes except line 21 and remove the extra slash and I'll merge. Thanks.

satterly:
Closing this in favour of #997

PR #997, opened 2019-06-07
Change Azure auth provider to use Active Directory v2.0

satterly:
Update AD endpoint as per #995 but also fix issuer URL verification and add extra error handling. Keep existing test case and rename to v1 and add a v2 test case.

Supercedes #995

/cc @erichester76

### DIFF
--- a/alerta/auth/oidc.py
+++ b/alerta/auth/oidc.py
@@ -18,7 +18,7 @@ from . import auth
 def get_oidc_configuration(app):
 
     OIDC_ISSUER_URL_BY_PROVIDER = {
-        'azure': 'https://sts.windows.net/{}/'.format(app.config['AZURE_TENANT']),
+        'azure': 'https://login.microsoftonline.com/{}/v2.0'.format(app.config['AZURE_TENANT']),
         'gitlab': app.config['GITLAB_URL'],
         'google': 'https://accounts.google.com',
         'keycloak': '{}/auth/realms/{}'.format(app.config['KEYCLOAK_URL'], app.config['KEYCLOAK_REALM'])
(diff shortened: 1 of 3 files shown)


alerta/auth/oidc.py
```diff
@@ -18,7 +18,7 @@
 def get_oidc_configuration(app):
 
     OIDC_ISSUER_URL_BY_PROVIDER = {
-        'azure': 'https://sts.windows.net/{}/'.format(app.config['AZURE_TENANT']),
+        'azure': 'https://login.microsoftonline.com/{}/v2.0'.format(app.config['AZURE_TENANT']),
```
satterly: This is the update to use v2.0 API endpoints.

alerta/auth/oidc.py
```diff
@@ -35,7 +35,7 @@ def get_oidc_configuration(app):
     except Exception as e:
         raise ApiError('Could not get OpenID configuration from well known URL: {}'.format(str(e)), 503)
 
-    if config['issuer'] != issuer_url:
+    if config['issuer'].format(tenantid=app.config['AZURE_TENANT']) != issuer_url:
```
satterly: This fixes the issuer URL verification for Azure which returns literally returns `{tenant}` in the issuer URL.

alerta/auth/oidc.py
```diff
@@ -73,6 +73,10 @@ def openid():
     r = requests.post(token_endpoint, data)
     token = r.json()
 
+    if 'error' in token:
+        error_text = token.get('error_description') or token['error']
+        raise ApiError(error_text)
+
```
satterly: This captures any errors returned by the identity provider and makes them visible in the web UI.

PR #1001, opened 2019-06-13
Allow LDAP auth with domain/username format

hamptons:
Adds the ability to login with <domain>\<username> when using the LDAP provider.

I tried to make all the changes in basic_ldap.py but unfortunately I had to edit user.py as well. If this is a sticking point it could be changed so that email login takes precedence, but I can only see it being a problem if somebody has a '\' in their username, which seems unlikely.

### DIFF
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -1,4 +1,5 @@
 
+import sys
 import ldap
 from flask import current_app, jsonify, request
 from flask_cors import cross_origin
@@ -17,12 +18,19 @@ from . import auth
 def login():
     # Retrieve required fields from client request
     try:
-        email = request.json.get('username', None) or request.json['email']
+        login = request.json.get('username', None) or request.json['email']
         password = request.json['password']
     except KeyError:
         raise ApiError("must supply 'username' and 'password'", 401)
 
-    username, domain = email.split('@')
+    email = ''
+    email_verified = False
+    if '\\' in login:
+        domain, username = login.split('\\')
+    else: 
+        username, domain = login.split('@')
+        email = login
+        email_verified = True
 
     # Validate LDAP domain
     if domain not in current_app.config['LDAP_DOMAINS']:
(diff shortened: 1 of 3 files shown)


alerta/auth/basic_ldap.py
```diff
@@ -17,12 +18,19 @@
...
 
-    username, domain = email.split('@')
+    email = ''
+    email_verified = False
+    if '\\' in login:
+        domain, username = login.split('\\')
+    else: 
+        username, domain = login.split('@')
+        email = login
+        email_verified = True
```
satterly: Vey minor, but for this I would put ...
```
    email = ''
    email_verified = False
```
after the `if` so that it looked symmetrical.

That is...
```
    if '\\' in login:
        domain, username = login.split('\\')
        email = ''
        email_verified = False
    else: 
        username, domain = login.split('@')
        email = login
        email_verified = True
```

alerta/auth/basic_ldap.py
```diff
@@ -40,11 +48,20 @@ def login():
     except Exception as e:
         raise ApiError(str(e), 500)
 
+    # Get email address from LDAP
+    if not email_verified:
+        try:
+            ldap_result = ldap_connection.search_s(userdn, ldap.SCOPE_SUBTREE,'(objectClass=*)',['mail'])
```
satterly: How generic is this attempt to get the email address? Would this work for most LDAP implementations, just some, or only yours?

alerta/auth/basic_ldap.py
```diff
@@ -40,11 +48,20 @@ def login():
...
+            email = ldap_result[0][1]['mail'][0].decode(sys.stdout.encoding)
```
satterly: Can we make the decode here more configurable? Looks like you can set a "encoding" config parameter that can be used for decoding responses. See https://ldap3.readthedocs.io/encoding.html#encoding

hamptons: ldap3 is a different package. I can't see any option to do similar in python-ldap - this is the most generic solution I can find

alerta/auth/basic_ldap.py
hamptons: I'm pretty sure it'll work for most. The 'mail' attribute seems to be the standard, can't find examples of anything else being used. Doesn't mean it'll contain a value of course.

I noticed that you need an email address to create an api key (and maybe elsewhere) which is why I put a placeholder in if the email can't be looked up. 

Can we not make the email address editable from the UI so that it can be updated if required?

alerta/auth/basic_ldap.py
satterly: Ok sorry about that. :+1+

alerta/auth/basic_ldap.py
satterly: You only need a login to create an API key so the email placeholder is not strictly necessary. And yes we could make email editable but I tend not to allow editing user attributes that are sourced from elsewhere otherwise you get discrepancies that are difficult to resolve.

satterly:
Thanks! 👍

PR #1004, opened 2019-06-19
Make oembed query database agnostic

### LINKED ISSUE #1003: Overview Page in Web-UI
Uptake of Alerta if that can be classed as a problem

Our operations dept is so reliant on Nagios/Vagvis as a dashboard/traffic light system that trying to move to alerta is proving problematic due to information overload looking at the landing alerts screen. (nagios alerts here are very spammy) 

It would be nice to see a single page overview of environments/systems in a traffic light form with nice big squares for each (i guess known) environment/system that is green for no alerts, yellow/red with maybe a count of alerts, clicking on these would simply lead to the current landing page with that environment or maybe filter already selected. 

as an extension to this, a configurable option by filters where a tile represents a filter or search would also be amazing. This would mean the operator can have searches for anything represented as a overview button and could mean that different groups could have overview pages of different things. As alerta is a great consolidator of alerts from everywhere this could be very handy.

making it myself using the alerta api/flask/python but I'd rather not as the new UI is looking nice.

**Additional context**
If you feel this is a good idea / easy to implement I'm happy to assist.

satterly:
Fixes #1003

### DIFF
--- a/alerta/views/oembed.py
+++ b/alerta/views/oembed.py
@@ -1,5 +1,5 @@
 
-from urllib.parse import urlparse
+from urllib.parse import parse_qs, urlparse
 
 from flask import current_app, jsonify, render_template, request
 from flask_cors import cross_origin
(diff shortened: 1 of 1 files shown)


seasurfpete:
did a quick test and now the page is responding back. quick fix Thanks, I will try it out to see if meets our needs now.

Issue #1009, opened 2019-06-27
Security: Authorization in OIDC and SAML not working as expected


maglo:
**Issue Summary**
If using `ALLOWED_OIDC_ROLES` as described in documentation in https://docs.alerta.io/en/latest/authentication.html#keycloak-oauth2, authorization (granting access based on the requirement to have the roles) does not work.

The user is allowed access even though the roles specified are not in the token.

**Environment**
- OS: Linux 
- API version: 7.2.10
- Deployment: Docker

- Database: MongoDB

- Server config:
```
[...]

PR #1010, opened 2019-06-27
Fix authorization logic for OIDC and SAML

maglo:
See #1009

Changes logic for oidc and saml authorization checks so that
the authorization fails if either of the authorization checks
fail.


satterly:
👍 Thanks!

PR #1022, opened 2019-07-01
Allowing for configurable SAML metadata and string casting Descriptor 

sherif84:
after migrating from 6.8 alerta to 7.2.11 , SAML authentication broke for us for with 2 issues :

1.  the default "saml_settings" in saml.py assumes the metadata is a "remote" url while our configuration uses a local idp.xml metadata 
```
SAML2_CONFIG = {
    'metadata': {
        'local': ['/etc/idp.xml']
    },
    'allow_unknown_attributes': True,
    'organization': {
        'name': [('orgainization','en')],
        'display_name': ['orgainization'],
        "url": [("https://alerta.com","en")],
    },
    "service": {
        "sp": {
            "allow_unsolicited": True,
        }
    },
    'key_file': '/etc/alerta_saml.key',
    'cert_file': '/etc/alerta_saml.cert'
}
```

2. ( after fixing issue 1 by overwriting the default metadata ) when accessing the endpoint /auth/saml/metadata.xml the following error is shown "EntityDescriptor  object is not iterable", it simply needed a string casting

### DIFF
--- a/alerta/auth/saml.py
+++ b/alerta/auth/saml.py
@@ -47,6 +47,9 @@ def get_saml2_config():
     if current_app.config['SAML2_ENTITY_ID']:
         saml_settings['entityid'] = current_app.config['SAML2_ENTITY_ID']
 
+    if current_app.config['SAML2_CONFIG'].get('metadata'):
+        saml_settings['metadata'] = current_app.config['SAML2_CONFIG']['metadata']
+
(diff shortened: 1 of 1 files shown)


satterly:
👍Thanks for reporting the issue and providing a fix. 🎉

sherif84:
glad to help :)

PR #1024, opened 2019-07-03
Update database backends to make  field updatable

bcanyelles:
The scenerio we need this change is:

* Enter an alert with event `exceptionAlert` as event type.
* For some reason the monitor detect a flapping and notify this event as `flappingAlert`.
* After that the alert is recovered from flapping state and notify it.
* Supose the next event is an execptionAlert event type again, the system should reflect it.


satterly:
You don't explain why. And the "why" is the most important part of a change request. Why are you changing the event type when you detect a flapping alert? Is there some downstream system that depends on this? Are you displaying it in the web UI? Or some other reason?

bcanyelles:
Hi and thanks in advance for the support,

We have several reasons to request this:

* First, alarms state changes must be showed in UI so the operators can acknowledge flapping state in a certain moment.
* Second, we think flapping status is a particular state for an alarm as it gives more information and it is easier to catch than having alerts up and down constantly.
* Besides that, if this change is accepted, there is no need of deploying a plugin to handle it anymore.

Best Regards

satterly:
Can you describe what the plugin that you are currently using for this does? And why isn't that a suitable solution?

bcanyelles:
I haven't implemented yet, but the behaviour would be like explained before: set an attribute that describes the event type received is flapping or not.
I think this field represents this concept and the implementation on the core is easier and more efficient than in additional plugin.

satterly:
I disagree. You are trying to encode two pieces of information in to the same field. ie. the type of the alert and whether it is flapping or not. Then to use the information you need to parse the field to find out whether it is flapping or not.

I think it's more explicit (and therefore easier to understand and maintain) to use a custom attribute to keep track of flapping status.

bcanyelles:
Thanks for your arguments.
I will consider them and implement this change at the plugin level.
Thank you!

PR #1035, opened 2019-07-26
Allow LDAP servers to use a self signed certificate

slyngshede:
Sometimes internal LDAP servers will use a self signed certificate. By default python-ldap will connections to LDAP servers with a self signed certificate. 

This allow the configuration of the python-ldap setting via Alertas configuration.

### DIFF
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -17,7 +17,7 @@ from . import auth
 @cross_origin(supports_credentials=True)
 def login():
     # Allow LDAP server to use a self signed certificate
-    if current_app.config['LDAP_TLS_ALLOW_SELFSIGNED']:
+    if current_app.config['LDAP_ALLOW_SELF_SIGNED_CERT']:
         ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_ALLOW)
 
     # Retrieve required fields from client request
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -16,6 +16,10 @@ from . import auth
 @auth.route('/auth/login', methods=['OPTIONS', 'POST'])
 @cross_origin(supports_credentials=True)
 def login():
+    # Allow LDAP server to use a self signed certificate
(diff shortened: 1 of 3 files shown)


alerta/auth/basic_ldap.py
```diff
@@ -16,6 +16,10 @@
 @auth.route('/auth/login', methods=['OPTIONS', 'POST'])
 @cross_origin(supports_credentials=True)
 def login():
+    # Allow LDAP server to use a self signed certificate
+    if current_app.config['LDAP_TLS_ALLOW_SELFSIGNED']:
```
satterly: Can you please rename `LDAP_TLS_ALLOW_SELFSIGNED ` to `LDAP_ALLOW_SELF_SIGNED_CERT`?

alerta/settings.py
```diff
@@ -87,6 +87,7 @@
 LDAP_DOMAINS = {}  # type: Dict[str, str]
 LDAP_DOMAINS_GROUP = {}  # type: Dict[str, str]
 LDAP_DOMAINS_BASEDN = {}  # type: Dict[str, str]
+LDAP_TLS_ALLOW_SELFSIGNED = False
```
satterly: Rename as per above.

alerta/utils/config.py
```diff
@@ -88,6 +88,9 @@ def get_user_config():
         if 'ALLOWED_GITLAB_GROUPS' in os.environ:
             config['ALLOWED_OIDC_ROLES'] = os.environ['ALLOWED_GITLAB_GROUPS'].split(',')
 
+        if 'LDAP_TLS_ALLOW_SELFSIGNED' in os.environ:
+            config['LDAP_TLS_ALLOW_SELFSIGNED'] = True if os.environ['LDAP_TLS_ALLOW_SELFSIGNED'] == 'True' else False
+
```
satterly: Please remove this as it's not needed. No other LDAP settings can be set by environment variable.

slyngshede: I really wanted it to be an environment variable for some reason. It's deleted.

alerta/auth/basic_ldap.py
slyngshede: Renamed.

slyngshede:
I'm a little unsure if I updated the pull request correctly.

PR #1036, opened 2019-07-31
telegram webhook: return error if alert_id is not found.

jjh74:
return error if Alert.find_by_id doesn't find alert_id (otherwise alert.set_status / alert.untag fail: 'NoneType' object has no attribute 'set_status').
(Discussed in https://github.com/alerta/alerta-contrib/issues/289).

(Note: I didn't test the bogus-id-this-shouldnt-exist test, I hope the test works / is ok).


satterly:
Thanks! 👍

PR #1039, opened 2019-08-02
X-API-Key and Authorization: Bearer headers.

jjh74:
To use X-API-Key header with msteams(webhook) we need to look for X-API-Key if Authorization header doesn't have key.
(this was briefly discussed in https://github.com/alerta/alerta-contrib/issues/280)

(Note: Changing L32 to if not key ... doesn't work, because if request has valid Authorization: Key then L35/36 (else block) always overwrites key variable).

### DIFF
--- a/alerta/auth/decorators.py
+++ b/alerta/auth/decorators.py
@@ -24,7 +24,7 @@ def permission(scope=None):
         def wrapped(*args, **kwargs):
 
             # API Key (Authorization: Key <key>)
-            if 'Authorization' in request.headers:
+            if 'Authorization' in request.headers and request.headers['Authorization'].find('Key ', 0, 4) == 0:
                 auth_header = request.headers['Authorization']
                 m = re.match(r'Key (\S+)', auth_header)
                 key = m.group(1) if m else None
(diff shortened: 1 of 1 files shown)


alerta/auth/decorators.py
```diff
@@ -24,7 +24,7 @@ def decorated(f):
         def wrapped(*args, **kwargs):
 
             # API Key (Authorization: Key <key>)
-            if 'Authorization' in request.headers:
+            if 'Authorization' in request.headers and request.headers['Authorization'].find('Key ', 0, 4) == 0:
```
satterly: why not `... and 'Key' in and request.headers['Authorization']` ?

jjh74: Do you mean L27:
```python
if 'Authorization' in request.headers and 'Key' in request.headers['Authorization']
``` 
My thinking was that .find('Key ', 0, 4) matches from beginning and 'in' might match: Authorization: Bearer ..somebase64string...Key...morebase64 ?

satterly: what about `... and request.headers['Authorization'].startswith('Key')` ?

jjh74: startswith works and is more readable than find('..', 0, 4) == 0 (I'm python newbie and find was first substring I found). I'll update PR.

satterly: No problem. Thanks for updating.

satterly:
👍

PR #1042, opened 2019-08-05
Pass timeout to action process_action and take_action functions

fshaikh-eg:
added kwarg for timeout for process_action so it can be used in a plugin during the take_action phase.

The need is to be able to use a plugin to change the timeout value based on the attributes of the alerts.  Currently the timeout value needs to be set upon the initial call to the API but we need an ability to overwrite that using a plugin, especially when trying to use the UI.

### DIFF
--- a/alerta/utils/api.py
+++ b/alerta/utils/api.py
@@ -108,8 +108,8 @@ def process_action(alert: Alert, action: str, text: str, timeout: int) -> Tuple[
         if updated:
             try:
                 if len(updated) == 3:
-                    alert, action, text = updated   
-                elif len(updated) == 4: 
+                    alert, action, text = updated
+                elif len(updated) == 4:
                     alert, action, text, timeout = updated
                 else:
                     alert = updated
--- a/alerta/utils/api.py
+++ b/alerta/utils/api.py
@@ -107,12 +107,14 @@ def process_action(alert: Alert, action: str, text: str, timeout: int) -> Tuple[
                 logging.error("Error while running action plugin '{}': {}".format(plugin.name, str(e)))
         if updated:
             try:
-                try:
-                    alert, action, text = updated    #if timeout does not get returned, try default behavior
-                except Exception:
-                    alert, action, text, timeout = updated #Try to see if timeout gets returned
+                if len(updated) == 3:
+                    alert, action, text = updated   
+                elif len(updated) == 4: 
+                    alert, action, text, timeout = updated
+                else:
+                    alert = updated
             except Exception:
-                alert = updated
+                logging.error("Error while running action plugin '{}': {}".format(plugin.name, str(e)))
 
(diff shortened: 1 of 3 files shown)


alerta/utils/api.py
```diff
@@ -87,16 +87,16 @@ def process_alert(alert: Alert) -> Alert:
     return alert
 
 
-def process_action(alert: Alert, action: str, text: str) -> Tuple[Alert, str, str]:
-
+def process_action(alert: Alert, action: str, text: str, time_out: int) -> Tuple[Alert, str, str, int]:
+    #named variable time_out so doesnt conflict with kwargs name
```
satterly: I don't understand why this is called "time_out" and not "timeout". Why would it conflict with `kwargs` name? 😕

alerta/utils/api.py
```diff
@@ -87,16 +87,16 @@ def process_alert(alert: Alert) -> Alert:
...
     wanted_plugins, wanted_config = plugins.routing(alert)
 
     updated = None
     for plugin in wanted_plugins:
         if alert.is_suppressed:
             break
         try:
-            updated = plugin.take_action(alert, action, text, config=wanted_config)
+            updated = plugin.take_action(alert, action, text, config=wanted_config, timeout=time_out)
```
satterly: Can you put "timeout" after "text" and before "config" ? The ordering is more consistent that way.

alerta/utils/api.py
fshaikh-eg: it just looked funny to me when it was timeout=timeout, but i guess there is not technical reason for it.

alerta/utils/api.py
fshaikh-eg: done.

alerta/utils/api.py
```diff
@@ -108,15 +107,18 @@ def process_action(alert: Alert, action: str, text: str) -> Tuple[Alert, str, st
                 logging.error("Error while running action plugin '{}': {}".format(plugin.name, str(e)))
         if updated:
             try:
-                alert, action, text = updated
+                try:
+                    alert, action, text = updated    #if timeout does not get returned, try default behavior
+                except Exception:
```
asherf: Ideally, we should catch TypeError here to catch unpacking errors and not a generic exception.

asherf: also might be cleaner, since `updated` is a tuple, to check the length and run logic based on that instead of doing try...except:
```
if len(updated) == 3:
  alert, action, text = updated   
elif len(updated) == 4: 
  alert, action, text, timeout = updated
else:
  alert = updated
```

fshaikh-eg: This is a much cleaner solution.  Thank you!

satterly:
Thanks. :+1:

PR #1044, opened 2019-08-05
Bump versions on pre-commit checks.

asherf:
Those bumps happened "automatically" when pre-commit ran locally, so I figured I'll push this update.


satterly:
:+1: Thanks

PR #1049, opened 2019-08-05
Mark regex strings as raw strings

asherf:
See https://lintlyci.github.io/Flake8Rules/rules/W605.html
Those strings cause deprecation warnings when running under python 3.6, flake 8 also complains about them:
Screenshot 2019-08-05 06 29 41


asherf:
this needs some more work... tests seem to fail with this change.

PR #1050, opened 2019-08-05
Mark regex strings as raw strings, properly escape strings

asherf:
See https://lintlyci.github.io/Flake8Rules/rules/W605.html
from https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior:
>A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases. (Contributed by Emanuel Barry in bpo-27364.)

Those strings cause deprecation warnings when running under python 3.6, flake 8 also complains about them:
Screenshot 2019-08-05 06 29 41


asherf:
ping.

satterly:
pong. 👍

PR #1054, opened 2019-08-10
Raven is deprecated, switch to sentry-sdk

asherf:
https://github.com/getsentry/raven-python#raven---sentry-for-python

>This SDK is being phased out for Sentry-Python.

https://docs.sentry.io/platforms/python/flask/


asherf:
needs some more work... will re-open when I get tests passing

PR #1056, opened 2019-08-10
Raven is deprecated, switch to sentry-sdk

asherf:
instead of https://github.com/alerta/alerta/pull/1054:
https://github.com/getsentry/raven-python#raven---sentry-for-python

>This SDK is being phased out for Sentry-Python.

https://docs.sentry.io/platforms/python/flask/

------------

The change in the custom webhook code: when debugging the previous [test failure](https://travis-ci.org/alerta/alerta/jobs/570122094) I noticed that when running with the new sentry sdk calling get_data hits a different code path:
https://github.com/pallets/werkzeug/blob/f753a326343cf5d163720f0d954d62aba3c5618c/src/werkzeug/wrappers/base_request.py#L451 
the `_cached_data` is populated (with `'say=Hi&to=Mom'`) and so this one takes, and that is the string that ends up getting passed to payload (to the webhook.incoming method)
with the old version (raven) get_data() returns an empty string and so we call request.form and that one returns the required result.
By looking at the sentry-sdk source code, you can see that the flask integration calls call of those methods to try and collect as much data on the request in case of an error, which causes _cached_data to be populated.
I have run this change (w/o the sentry upgrade) and all the tests seem to pass.
Moreover, it seems that we should prefer structured data (form) before resorting to raw data (string)

### DIFF
--- a/alerta/app.py
+++ b/alerta/app.py
@@ -35,7 +35,7 @@ key_helper = ApiKeyHelper()
 db = Database()
 qb = QueryBuilder()
 # Sentry will grab DSN from SENTRY_DSN environment variable.
-sentry_sdk.init(None, integrations=[FlaskIntegration()])
+sentry_sdk.init(integrations=[FlaskIntegration()])
 
 mailer = Mailer()
 plugins = Plugins()
--- a/alerta/app.py
+++ b/alerta/app.py
@@ -1,9 +1,10 @@
 from typing import Any, Dict
 
+import sentry_sdk
 from flask import Flask
 from flask_compress import Compress
 from flask_cors import CORS
-from raven.contrib.flask import Sentry
+from sentry_sdk.integrations.flask import FlaskIntegration
 from werkzeug.contrib.fixers import ProxyFix
 
 from alerta.database.base import Database, QueryBuilder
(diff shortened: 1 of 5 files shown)


alerta/app.py
```diff
@@ -33,13 +34,15 @@
 
 db = Database()
 qb = QueryBuilder()
-sentry = Sentry()
+# Sentry will grab DSN from SENTRY_DSN environment variable.
+sentry_sdk.init(None, integrations=[FlaskIntegration()])
```
satterly: Can you remove this `None`?

alerta/webhooks/custom.py
```diff
@@ -24,7 +24,7 @@ def custom(webhook):
     try:
         rv = custom_webhooks.webhooks[webhook].incoming(
             query_string=request.args,
-            payload=request.get_json() or request.get_data(as_text=True) or request.form
+            payload=request.get_json() or request.form or request.get_data(as_text=True)
```
satterly: 👍 Thanks for the explanation in the description.

satterly:
👍 Thanks!

alerta/app.py
asherf: Yes. Sorry about that, for some reason I thought this was a positional argument

PR #1057, opened 2019-08-10
run autopep8 fix

asherf:
pre-commit keeps running autopep8 when commiting changes which creates some unrelated changes in the commits/PRs
I figured we can just let it run once on the entire code base, fix whatever it wants to fix and be done with it.


satterly:
Fair enough. I was getting tired of undoing the extra spaces around equals signs. Sigh. 👍

PR #1058, opened 2019-08-10
iSort fixes

asherf:
Along the lines of https://github.com/alerta/alerta/pull/1057 but actually dangerous because isort tries to change import order in some places where it is an undesired outcome.
so this PR fixes sorting where it makes sense and tells isort to ignore where it doesn't.


satterly:
Thanks. This was very annoying. 👍

PR #1061, opened 2019-08-10
Fix flake W504 issues and ignore flake W503

asherf:
[W503](https://lintlyci.github.io/Flake8Rules/rules/W503.html) is not considered a best practice
[W504](https://lintlyci.github.io/Flake8Rules/rules/W504.html) does, so fix those issues that show up in precommit


satterly:
👍

PR #1063, opened 2019-08-12
Switch to using pytest

asherf:
nose is [effectively dead](https://nose.readthedocs.io/en/latest/#note-to-users) and even [nose2 docs](https://github.com/nose-devs/nose2#nose2-vs-pytest) says to use pytest
so I think it is a good idea to switch to using pytest


satterly:
Good suggestion. Thanks. :+1:

PR #1064, opened 2019-08-13
Add coveralls - code coverage service

asherf:
You will need to create a [coveralls](https://github.com/marketplace/coveralls) account (free for open source) and add the repo to coveralls.
then add the repo token to [travis](https://travis-ci.org/alerta/alerta/settings)
Screenshot 2019-08-13 09 14 56

Feel free to ping me in gitter, I am in the alerta chat there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ install:
   - pip install .
 
 script:
+  - pip install coveralls
   - python -m mypy alerta/
   - pre-commit run -a
   - pytest --cov=alerta
+  - coveralls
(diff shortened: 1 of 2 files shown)


satterly:
Can you put coveralls in `requirements-dev.txt`?

asherf:
@satterly doesn't make a lot of sense to install it when developing since all that it does it to upload a coverage report to coveralls, which we probably only want to do from CI.
the logic to calculate coverage as test run is part of the pytest-cov package which is already in  `requirements-dev.txt`.

PR #1067, opened 2019-08-14
Fix monogdb deprecations

asherf:
https://api.mongodb.com/python/3.7.2/api/pymongo/collection.html?highlight=delete_many#pymongo.collection.Collection.remove


satterly:
👍 Thanks.

PR #1087, opened 2019-10-02
Changing order of hearbeat status evaluation

joaorebelo-ar:
Changing order of hearbeat status evaluation to prioritize stale against slow.


satterly:
Thanks 👍

PR #1095, opened 2019-10-08
Import ProxyFix from werkzeug.middleware.proxy_fix

jjh74:
(werkzeug.contrib.fixers is deprecated: https://werkzeug.palletsprojects.com/en/0.15.x/contrib/fixers/)


satterly:
Thanks. I'll fix the broken build in a following PR. 👍

PR #1098, opened 2019-10-12
utils/audit.py:get_redacted_data: check that data is defined

jjh74:
otherwise (at lease heartbeat delete) can fail with:
File "alerta/utils/audit.py", line 86, in get_redacted_data
if 'password' in data:
TypeError: argument of type 'NoneType' is not iterable

### DIFF
--- a/alerta/utils/audit.py
+++ b/alerta/utils/audit.py
@@ -82,7 +82,7 @@ class AuditTrail:
 
         def get_redacted_data(r):
             data = r.get_json()
-            if app.config['AUDIT_LOG_REDACT']:
+            if data and app.config['AUDIT_LOG_REDACT']:
                 if 'password' in data:
                     data['password'] = '[REDACTED]'
             return json.dumps(data)


jjh74:
I found this when looking why alerta-cli fails to delete heartbeat: alerta heartbeat -D <id> sends DELETE request with Content-Length: 0 / Content-Type: application/json.
Alerta will fails this request:
{
  "code": 400, 
  "errors": [
    "Failed to decode JSON object: Expecting value: line 1 column 1 (char 0)"
  ], 
  "message": "400 Bad Request: Failed to decode JSON object: Expecting value: line
 1 column 1 (char 0)", 
  "status": "error"
}

Same happens with curl: curl -v -XDELETE -H 'Authorization: Key keyhere...' -H "Content-Length: 0" -H "Content-Type: application/json" https://hostname.example.org/api/heartbeat/<id>

(Sending the same request without Content-Length/Type works)

satterly:
👍 Thanks!

PR #1111, opened 2019-11-11
#1079 breaks changing user passwords.

jjh74:
https://github.com/alerta/alerta/pull/1079 breaks changing user passwords with internal authentication (passwords are set to [REDACTED]).
This commit moves user.update / user.update_attributes before
write_audit_trail, so password is set before get_redacted_data sets data['password'] to '[REDACTED]'.


satterly:
Thanks. 👍

PR #1120, opened 2020-01-02
Fix for error in the id option of the history command

naureensaba:
Fixed the error received while using the 'id' option of the 'history' command.


satterly:
Thanks for the PR but the fix was a little more complicated than it looked. I've looked at it myself and have created a new PR.

PR #1126, opened 2020-01-20
Fix admin role to scope lookup

satterly:
/cc @burakksglu


burakksglu:
Thanks! Nice PR. @satterly

burakksglu:
I tested it @satterly. The fix works as expected. Thanks for the effort!

PR #1127, opened 2020-01-22
Added guest user with the scope of "read.alerts"

burakksglu:
_**Is your request related to a problem? Please describe.**_

It is, the problem is when you setup several alerta instances, you have to add guest user with read only role named at the start manually.

_**Describe the solution you'd like**_

We can create a guest user with a read only role(read.alerts only) at initial setup of docker image or alerta app.

_**Describe solutions you've considered**_

We created the role by hand which was not very convenient for multiple setups.

I actually tested this with some change to the "permission.py" file. I created an extra user with the scope of 'read:alerts'. So adding few lines of code we can actually have 3 IdP roles ready at start.

- Admin(Scope: Admin)
- User(Scope: Write,Read)
- Guest(Scope: Read.alerts)

Extra context & screenshots:
Screenshot from 2020-01-21 15-48-22
image

### DIFF
--- a/alerta/views/permissions.py
+++ b/alerta/views/permissions.py
@@ -75,6 +75,11 @@ def list_perms():
         scopes=current_app.config['USER_DEFAULT_SCOPES']
     )
 
+    guest_perm = Permission(
+        match='guest',
+        scopes=[Scope.read_alerts]
+    )
+
     # add system-defined roles 'admin' and 'user'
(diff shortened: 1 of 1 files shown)


satterly:
Thanks for that but there's a little more to it than this. I'll review in more detail and suggest improvements when I get some time.

satterly:
Closing in favour of #1131

Issue #891, opened 2019-03-22
Ability to create backdated alerts matching expired blackouts


Anderen2:
**Is your feature request related to a problem? Please describe.**
We sometimes experience some race-conditions with our alerting and blackouts. 
This happens when events generated by downstream systems (eg. Telegraf/Kapactor, etc) get delayed to Alerta, and is raised after a blackout has automatically expired or been deleted. 

**Describe the solution you'd like**
The ability to create back-dated alerts that matches expired blackouts. 
Likely requires change in https://github.com/alerta/alerta/blob/master/alerta/database/backends/postgres/base.py#L643 to take into account that createTime may be in the past. 

**Describe alternatives you've considered**
A temporary workaround for this is to create blackouts with a longer expiry that takes the delay into account.

satterly:
So are you suggesting instead of this...
```
WHERE start_time <= NOW() at time zone 'utc' AND end_time > NOW() at time zone 'utc'
```
it should be something like this ...
```
WHERE start_time <= alert.createTime AND end_time > alert.createTime
```

Have you tried/tested this?

PR #1134, opened 2020-01-29
Use lastReceiveTime and createTime for blackouts

### LINKED ISSUE #1132: Blackout timerange check uses create time, not a current one
**Issue Summary**

The current code changes state of the alert to blackout only if a create time is in the range of blackout. That's not what users expect. They expect it to transition to blackout if blackout is active now.

**Environment**
- OS: Linux
- API version: 7.4.1
- Deployment: self-hosted
- For self-hosted, WSGI environment: nginx/uwsgi

- Database: [eg. MongoDB, Postgres]

- Server config:
  Auth enabled? Yes
  Auth provider? Basic
  Customer views? No
  (or provide the `/config` endpoint output)

- web UI version: any
- CLI version: any

**To Reproduce**
1. Create a blackout for a resource which already has open alert(s).
2. Alert(s) remain open regardless of an active blackout and received duplicates.

**Expected behavior**
I'd expect all affected alerts changing status to blackout regardless how old these are.

**Additional context**
We are using alertmanager as a main source of alerts an it's periodically refreshes alerts. All alerts have a timeout set.

satterly:
Changing `createTime` to `receiveTime` would have just reintroduced the problem fixed by #891. The solution was to use both `createTime` (which is what it uses currently) for newly created alerts during a blackout period and `receiveTime` (or more correctly, `lastReceiveTime`) for alert duplicates created during a blackout period.

Fixes #1132

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -891,11 +891,19 @@ class Backend(Database):
 
     def is_blackout_period(self, alert):
         query = dict()
-        query['startTime'] = {'$lte': alert.create_time}
-        query['endTime'] = {'$gt': alert.create_time}
-
         query['environment'] = alert.environment
-        query['$and'] = [{'$or': [
+        query['$and'] = [
+            {'$or': [
+                {'$and': [
+                    {'startTime': {'$lte': alert.create_time}},
+                    {'endTime': {'$gt': alert.create_time}}
+                ]},
+                {'$and': [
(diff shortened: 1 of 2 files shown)


satterly:
Closing this hack to blackout periods (which won't work) in favour of fixing the root cause which are integrations that incorrectly (unnecessarily) set the create time to a fixed alarm "start time" which has no meaning to Alerta yet makes suppressing these alerts using blackout periods very difficult, if not impossible. See #1132

PR #1143, opened 2020-02-13
Allow to log audited alert data as JSON object

escoand:
I would like to log the alert data in audit as JSON object. So it could be used directly instead of parsing the data attribute again.


satterly:
Thanks 👍

Issue #1073, opened 2019-09-09
Multi-level OIDC_ROLE_CLAIM and OIDC_GROUP_CLAIM


escoand:
**Is your feature request related to a problem? Please describe.**
I use Keycloak as OIDC provider and want to use OIDC_GROUP_CLAIM and OIDC_ROLE_CLAIM. But Keycloak reponds with a multi-level claim structure which is currently not possible to use.
```json
{
  "name": "Me",
  "email": "me@example.com",
  "realm_access": {
    "roles": [
      "just_some_generic_roles"
    ]
  },
  "resource_access": {
    "RANDOM_OAUTH2_CLIENT_ID": {
      "roles": [
        "my_role_1",
        "my_role_2"
      ]
    }
  }
}
```

**Describe the solution you'd like**
According to the data above I need something like `resource_access/RANDOM_OAUTH2_CLIENT_ID/roles` as `OIDC_GROUP_CLAIM` and `OIDC_ROLE_CLAIM`.
This leads to `data.get('resource_access').get('RANDOM_OAUTH2_CLIENT_ID').get('roles')` at https://github.com/alerta/alerta/blob/master/alerta/auth/oidc.py#L123.

**Describe alternatives you've considered**
I'm not the supporter of the Keycloak server so it's impossible to have the alternative on the server side.

escoand:
I've found a relatively simple solution for it: https://stackoverflow.com/a/36131992
```python
from functools import reduce
reduce(lambda d, k: d.get(k) if d else None, ['resource_access', 'RANDOM_OAUTH2_CLIENT_ID', 'roles'], data)
# ['my_role_1', 'my_role_2']
```

satterly:
Are you sure your Keycloak server is configured correctly? See test case for expected payload format...
```
        {
          "sub": "4eeaa122-08d8-47bc-b6d0-010481c055c5",
          "email_verified": true,
          "roles": [
            "create-realm",
            "devops",
            "alerta-project",
            "admin"
          ],
          "name": "Nicholas Satterly",
          "preferred_username": "nsatterl",
          "given_name": "Nicholas",
          "family_name": "Satterly",
          "email": "nick@alerta.dev"
        }
```
https://github.com/alerta/alerta/blob/master/tests/test_providers.py#L1241

escoand:
Don't know if this is correct or not. But it's what the server gives me.

satterly:
You need to create a role and group mapper in the keycloak server configuration. For example ...
[...]

PR #1144, opened 2020-02-13
Read default keycloak realm and resouce based roles and groups

escoand:
Implemented the ability to read the default keycloak realm and resource based roles and groups. Yes, there is a documentation how to configure right but in my case I'm not able to change the configuration because it's a central provided service. Additionally this is the default keycloak configuration and nothing special for my deployment.

```json
{
  "name": "Me",
  "email": "me@example.com",
  "realm_access": {
    "roles": [
      "just_some_generic_roles"
    ]
  },
  "resource_access": {
    "RANDOM_OAUTH2_CLIENT_ID": {
      "roles": [
        "my_role_1",
        "my_role_2"
      ]
    }
  }
}
```

We already discussed this at https://github.com/alerta/alerta/issues/1073

This implementation uses the keycloak-gatekeeper (proxy service for keycloak) as a basis: https://github.com/keycloak/keycloak-gatekeeper/blob/master/user_context.go#L56-L76

### DIFF
--- a/alerta/auth/oidc.py
+++ b/alerta/auth/oidc.py
@@ -127,8 +127,8 @@ def openid():
             group_claim: access_token.get('realm_access', {}).get(group_claim, []) +
                          access_token.get('resource_access', {}).get(client_id, {}).get(group_claim, []),
         }
-    except Exception:
-        current_app.logger.warning('No access token in OpenID Connect token response.')
+    except Exception as err:
+        current_app.logger.warning('No access token in OpenID Connect token response: %s', err)
         keycloak_claims = {}
 
     try:
--- a/alerta/auth/oidc.py
+++ b/alerta/auth/oidc.py
@@ -148,8 +148,8 @@ def openid():
     picture = userinfo.get('picture') or id_token.get('picture')
 
     custom_claims = {
-        role_claim: keycloak_claims.get(role_claim) or userinfo.get(role_claim) or id_token.get(role_claim) or [],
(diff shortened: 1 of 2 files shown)


escoand:
In the mentioned issue we also discussed about roles in the access_token and you [stated](https://github.com/alerta/alerta/issues/1073#issuecomment-530981002):
> ID tokens are always JWTs but access tokens are not defined in the OpenID Connect or OAuth2 spec as JWTs. Most access tokens are opaque strings that have no meaning. Keycloak just happens to use JWTs, many others don't.

I now stumbled over a chapter of the [keycloak docs](https://www.keycloak.org/docs/latest/securing_apps/#openid-connect-2) which is talking about the same:
> There are really two types of use cases when using OIDC. The first is an application that asks the Keycloak server to authenticate a user for them. After a successful login, the application will receive an identity token and an access token. The identity token contains information about the user such as username, email, and other profile information. The access token is digitally signed by the realm and contains access information (like user role mappings) that the application can use to determine what resources the user is allowed to access on the application.

This sound to me like having the roles in the access_token is the default for Keycloak and having in id_token is only a possiblity.

escoand:
@satterly I've added the tests but are failing and don't understand why, can you explain?
https://travis-ci.org/alerta/alerta/jobs/655749947#L698

satterly:
Tests are broken
```
tests/test_providers.py ..F.FFFF.                                        [ 55%]
```

PR #1145, opened 2020-02-13
Ability to send HTML mails

escoand:
I needed to send HTML mails from a custom plugin. I use the build-in mailer but need to patch it, to allow HTML mime types.


satterly:
:+1:

PR #1154, opened 2020-02-20
Use OIDC and user roles

escoand:
When using OIDC authentication the defined user roles are not used anymore. Was this by accident or intentionally? This PR changes the behavior to use both role sources.


satterly:
The last time this code was changed it looked like this ...
```
roles = userinfo.get(current_app.config['OIDC_CUSTOM_CLAIM'], ['user'])
```

satterly:
👍

PR #1206, opened 2020-04-21
LDAP login improvements - Adding ldap search query.

### LINKED ISSUE #1172: LDAP Login configuration improvements
Trying to adapt and configure the LDAP authentication system, I have faced many problems because the way LDAP is configured is very strict and it doesn't let having a dynamic configuration.

The first issue I have faced is that I cannot determine the DN on a configuration file. The DN in my LDAP system doesn't contain the username (sAMAccountName field is the field which contain the username). So I can perform login if I trick the system, I set as DN under *LDAP_DOMAINS* variable as 'domain\username', but the Alerta LDAP process understand this as userDN, which is not the case. This userDN then cannot be used in the Groups filter as this is not a valid LDAP DN

So, in order to solve this problem, I would suggest to follow the structure that most of the applications out there uses:
* Define a User search filter per domain (to keep compatibility with the current system, I would create a new variable. If User search filter is defined for the current domain, use it, otherwise, follow the current system to get the DN to attempt the login)
* When a user is trying to log in
  * If user search filter exist
    * Search the user using the provided User Search filter for the current domain
    * If one user is found
      * Set the DN as the one found
    * if more than one user is found
      * Except: Search query is bad defined
   * else
     * Set the DN as the one found in *LDAP_DOMAINS* variable
   * Perform login with the set DN
     * If login successful
       * Perform group search using Group Search filter defined for the current domain

As you can see, this procedure is quite similar but the only difference is that the DN structure is not defined in a configuration file, instead, the search query to get the DN is defined.

With the User search query, the system is more flexible and the login can be performed using a valid LDAP field, in this case, you can use a valid email address as well if you want to use email address instead of username (domain\email or domain\username)

If you think this solution can be implemented and added to Alertaio, I can perform a PR. Let me know and I will do

frasy:
LDAP login improvements - Adding ldap search query. Instead of providing a fixed DN, a ldap search query can be provided to perform the user query and locate the DN

New properties added:
* LDAP_BIND_USERNAME: User DN to execute the LDAP Search query
* LDAP_BIND_PASSWORD: User DN password to execute the LDAP Search query
* LDAP_DOMAINS_USER_BASEDN: Base DN where to search for users, this variable has been created to have a better performance in the search user filter. If it is not defined, the Base DN will be used and the whole tree will be queried.
* LDAP_DOMAINS_GROUP_BASEDN: Base DN where to search for groups, this variable has been created to have a better performance in the search user filter. If it is not defined, the Base DN will be used and the whole tree will be queried.
* LDAP_DOMAINS_SEARCH_QUERY: Query to locate the user DN in the LDAP server. The query must return one entry. If there is no query defined for the current domain, the old system will happen and the static DN will be searched

Fixes #1172

### DIFF
--- a/alerta/auth/__init__.py
+++ b/alerta/auth/__init__.py
@@ -36,7 +36,7 @@ class AuthBlueprint(Blueprint):
 auth = AuthBlueprint('auth', __name__)
 
 
-from . import github, logout, oidc, userinfo   # noqa isort:skip
+from . import github, login, logout, oidc, userinfo   # noqa isort:skip
 
 
 @auth.before_request
--- a/alerta/auth/basic.py
+++ b/alerta/auth/basic.py
@@ -63,8 +63,6 @@ def signup():
     return jsonify(token=token.tokenize)
 
 
-@auth.route('/auth/login', methods=['OPTIONS', 'POST'])
-@cross_origin(supports_credentials=True)
 def login():
     # lookup user from username/email
     try:
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -2,7 +2,6 @@ import sys
 
 import ldap  # pylint: disable=import-error
 from flask import current_app, jsonify, request
-from flask_cors import cross_origin
 
 from alerta.auth.utils import create_token, get_customers
 from alerta.exceptions import ApiError
@@ -10,11 +9,7 @@ from alerta.models.permission import Permission
 from alerta.models.user import User
 from alerta.utils.audit import auth_audit_trail
 
-from . import auth
 
-
-@auth.route('/auth/login', methods=['OPTIONS', 'POST'])
-@cross_origin(supports_credentials=True)
 def login():
     # Allow LDAP server to use a self signed certificate
     if current_app.config['LDAP_ALLOW_SELF_SIGNED_CERT']:
@@ -40,15 +35,61 @@ def login():
         raise ApiError('expected username with domain', 401)
 
     # Validate LDAP domain
-    if domain not in current_app.config['LDAP_DOMAINS']:
+    if domain not in current_app.config['LDAP_DOMAINS'] and \
+       domain not in current_app.config['LDAP_DOMAINS_SEARCH_QUERY']:
         raise ApiError('unauthorized domain', 403)
 
-    userdn = current_app.config['LDAP_DOMAINS'][domain] % username
-
-    # Attempt LDAP AUTH
+    # Initialise ldap connection
     try:
         trace_level = 2 if current_app.debug else 0
         ldap_connection = ldap.initialize(current_app.config['LDAP_URL'], trace_level=trace_level)
+    except Exception as e:
+        raise ApiError(str(e), 500)
+
+    # If user search filter exist
+    #   Search the user using the provided User Search filter for the current domain
+    #   If one user is found
+    #       Set the DN as the one found
+    #       Set email retreived from AD
+    #   If more than one user is found
+    #       Except: Search query is bad defined
+    # Else
+    #   Set the DN as the one found in LDAP_DOMAINS variable
+    domain_search_query = current_app.config.get('LDAP_DOMAINS_SEARCH_QUERY', {})
+    base_dns = current_app.config.get('LDAP_DOMAINS_BASEDN', {})
+    user_base_dn = current_app.config.get('LDAP_DOMAINS_USER_BASEDN', {})
+    if domain in domain_search_query:
+        ldap_bind_username = current_app.config.get('LDAP_BIND_USERNAME', '')
+        ldap_bind_password = current_app.config.get('LDAP_BIND_PASSWORD', '')
+
+        try:
+            ldap_connection.simple_bind_s(ldap_bind_username, ldap_bind_password)
+        except ldap.INVALID_CREDENTIALS:
+            raise ApiError('invalid ldap bind username or password', 401)
+
+        ldap_users = [(_dn, user) for _dn, user in ldap_connection.search_s(
+            base_dns[domain] if user_base_dn.get(domain) is None else user_base_dn[domain],
+            ldap.SCOPE_SUBTREE,
+            domain_search_query[domain].format(username=username, email=email),
+            ['mail']
+        ) if _dn is not None]
+
+        if len(ldap_users) > 1:
+            raise ApiError('invalid search query for domain "{}"'.format(domain), 500)
+        elif len(ldap_users) == 0:
+            raise ApiError('invalid username or password', 401)
+
+        for _dn, _email in ldap_users:
+            userdn = _dn
+            email_attr = _email.get('mail')
+            if email_attr is not None:
+                email = email_attr[0].decode(sys.stdout.encoding)
+                email_verified = True
+    else:
+        userdn = current_app.config['LDAP_DOMAINS'][domain] % username
+
+    # Attempt LDAP AUTH
+    try:
         ldap_connection.simple_bind_s(userdn, password)
     except ldap.INVALID_CREDENTIALS:
         raise ApiError('invalid username or password', 401)
@@ -78,17 +119,18 @@ def login():
     groups = list()
     try:
         groups_filters = current_app.config.get('LDAP_DOMAINS_GROUP', {})
-        base_dns = current_app.config.get('LDAP_DOMAINS_BASEDN', {})
-        if domain in groups_filters and domain in base_dns:
+        groups_base_dn = current_app.config.get('LDAP_DOMAINS_GROUP_BASEDN', {})
+        if domain in groups_filters and (domain in base_dns or domain in groups_base_dn):
             resultID = ldap_connection.search(
-                base_dns[domain],
+                base_dns[domain] if groups_base_dn.get(domain) is None else groups_base_dn[domain],
                 ldap.SCOPE_SUBTREE,
                 groups_filters[domain].format(username=username, email=email, userdn=userdn),
                 ['cn']
             )
             resultTypes, results = ldap_connection.result(resultID)
             for _dn, attributes in results:
-                groups.append(attributes['cn'][0].decode('utf-8'))
+                if _dn is not None:
+                    groups.append(attributes['cn'][0].decode('utf-8'))
     except ldap.LDAPError as e:
         raise ApiError(str(e), 500)
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,6 +4,7 @@ pylint==2.5.2
 pytest-cov
 pytest>=5.4.3
 python-dotenv
+python-ldap
 requests_mock
 twine
 wheel
--- a/alerta/auth/login.py
+++ b/alerta/auth/login.py
@@ -1,0 +1,15 @@
+from flask import current_app
+from flask_cors import cross_origin
+
+from . import auth
+
+
+@auth.route('/auth/login', methods=['OPTIONS', 'POST'])
+@cross_origin(supports_credentials=True)
+def login():
+    if current_app.config['AUTH_PROVIDER'] == 'ldap':
+        from . import basic_ldap
+        return basic_ldap.login()
+    else:
+        from . import basic
+        return basic.login()
--- a/alerta/settings.py
+++ b/alerta/settings.py
@@ -96,9 +96,14 @@ ALLOWED_GITLAB_GROUPS = None
 
 # BasicAuth using LDAP
 LDAP_URL = ''  # eg. ldap://localhost:389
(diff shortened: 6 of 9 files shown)


frasy:
fixes #1172

frasy:
@satterly could you have the change to look at this PR

satterly:
Hi Thanks for the PR. The problem I have with reviewing this PR is there are no tests that cover any of the code. This is not your fault. It was already like this. And as this is a significant change, and many people rely on LDAP for auth, I can't review it without testing it first. So either I have to write some tests or spin up an LDAP server myself and test it, or both. Finding the time to do this is going to take some time. Let me know if you have any suggestions on how we could progress this PR faster.

frasy:
Hi @satterly, I have managed to add some test for the LDAP module but I had to add a big change:
* At testing level, Python Unit Test only permits a single route for '/auth/login', so I have moved the route to another file called 'main_auth.py' (if you know how to avoid this, let me know because I wasn't able to do it)
* Regarding to pylintrc module, I have deactivated the ldap module check because the lint process failed with no reason. I guess that the failing is due to ldap module is half python-half C library

Have a look and let me know.

frasy:
@satterly Did you have any chance to have a look at this?

frasy:
Hi @satterly, do you think you can have a look at this any time soon. It will really help me having this feature in alerta so I can use proper queries for login.

satterly:
Can you provide an example LDIF file that you may have used to test and a corresponding `alertad.conf` file that uses your new config settings to authenticate against it? That would help enormously for my testing. If you don't have an example LDIF file you could use the one here https://spring.io/guides/gs/authenticating-ldap/ Thanks

alerta/auth/basic_ldap.py
```diff
@@ -2,20 +2,16 @@
...
 from alerta.models.permission import Permission
 from alerta.models.user import User
 from alerta.utils.audit import auth_audit_trail
 
-from . import auth
 
-
-@auth.route('/auth/login', methods=['OPTIONS', 'POST'])
-@cross_origin(supports_credentials=True)
 def login():
```
satterly, line 13: change this line to ...

    def ldap_login():

... and you can revert all the changes you had to do to make testing work.

>At testing level, Python Unit Test only permits a single route for '/auth/login', so I have moved the route to another file called 'main_auth.py' (if you know how to avoid this, let me know because I wasn't able to do it)

frasy: The change you propose, I did already try before pushing those changes.

The reason I had to do these changes is because if I run the test in an individual way, they run correctly, but If I run them along with the other cases, for some reason the context and the app seems to be already loaded and the LDAP endpoint is not hit.

As you can see in the next executions (with the change you proposed and discarding every change I did), somehow the context is persisted from one test to another
```
TOXENV="postgres tests/test_auth.py::AuthTestCase tests/test_auth_ldap.py::AuthLdapTestCase" make test
```
```
tests/test_auth.py .........
tests/test_auth_ldap.py .<Response streamed [401 UNAUTHORIZED]>
F<Response streamed [401 UNAUTHORIZED]>
F<Response streamed [401 UNAUTHORIZED]>
F<Response streamed [401 UNAUTHORIZED]>
F
```

```
TOXENV="postgres tests/test_auth_ldap.py::AuthLdapTestCase" make test
```
```
tests/test_auth_ldap.py .2020-06-09 10:18:36,011 alerta.app[33381]: [WARNING] LDAP LOGIN request_id=d262d1c9-5bfc-4ef9-bdc2-a7f4698573c4 ip=127.0.0.1
<Response streamed [200 OK]>
.2020-06-09 10:18:36,410 alerta.app[33381]: [WARNING] LDAP LOGIN request_id=0352eae9-57a2-4e5d-b1bc-f741364f2eca ip=127.0.0.1
<Response streamed [200 OK]>
.2020-06-09 10:18:36,795 alerta.app[33381]: [WARNING] LDAP LOGIN request_id=8a292aec-2d7a-4d2d-8fcc-455bd0a2b1f0 ip=127.0.0.1
<Response streamed [401 UNAUTHORIZED]>
.2020-06-09 10:18:37,207 alerta.app[33381]: [WARNING] LDAP LOGIN request_id=8a04a30a-8e3f-4748-bcf1-bf028942344c ip=127.0.0.1
<Response streamed [401 UNAUTHORIZED]>
.
```

The warning you see is a warning I added at the beginning of the login function in basic_ldap.py module and as you can see is only hit when the test case is run by itself.

I've been trying to avoid this case, I searched a way to remove the context to avoid those code changes, but I saw nothing it could help.

If you know a way to avoid that, I would really appreciate if you can tell me

satterly:
Try changing the `tox.ini` file to this...

```
[tox]
envlist = py{37}-{mongodb,postgres}-{basic,ldap}
skip_missing_interpreters=true

[testenv]
deps =
  pytest
  requests_mock
  mongodb: pymongo
  postgres: psycopg2
  ldap: python-ldap

whitelist_externals =
  createdb
  dropdb
  pytest

commands_pre =
  postgres: -createdb alerta
commands = pytest -s {posargs:tests}
commands_post =
  postgres: -dropdb alerta

#passenv = *
setenv =
  ALERTA_SVR_CONF_FILE =
  mongodb: DATABASE_URL = mongodb://127.0.0.1:27017/alerta
  postgres: DATABASE_URL = postgres:///alerta

```

frasy:
With the LDAP option in tox you are saying if the python-ldap module should be installed. That is ok for a single test, but if you run the whole set, we have the same problem as commented before:
```
TOXENV="postgres-ldap" make test
```
```
tests/test_actions.py ..
tests/test_aggregations.py .
tests/test_alerts.py ................
tests/test_auth.py .........
tests/test_auth_ldap.py .<Response streamed [401 UNAUTHORIZED]>
F<Response streamed [401 UNAUTHORIZED]>
F<Response streamed [401 UNAUTHORIZED]>
F<Response streamed [401 UNAUTHORIZED]>
F
tests/test_blackouts.py .......
tests/test_commands.py ....

...
```
```
TOXENV="postgres-basic" make test
  or
TOXENV="postgres" make test
```
```
___________________________________________________________________________________ ERROR collecting tests/test_auth_ldap.py ___________________________________________________________________________________
ImportError while importing test module '../tests/test_auth_ldap.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_auth_ldap.py:5: in <module>
    import ldap
E   ModuleNotFoundError: No module named 'ldap'
=========================================================================================== short test summary info ============================================================================================
ERROR tests/test_auth_ldap.py
```
```
TOXENV="postgres-ldap tests/test_auth_ldap.py::AuthLdapTestCase" make test
```
```
tests/test_auth_ldap.py .2020-06-10 07:45:51,793 alerta.app[8385]: [WARNING] LDAP LOGIN request_id=991d6edb-70bd-4cae-8e13-21faa55e8b2a ip=127.0.0.1
<Response streamed [200 OK]>
.2020-06-10 07:45:52,028 alerta.app[8385]: [WARNING] LDAP LOGIN request_id=80c37094-7033-488f-a12a-7f78083dd2b8 ip=127.0.0.1
<Response streamed [200 OK]>
.2020-06-10 07:45:52,288 alerta.app[8385]: [WARNING] LDAP LOGIN request_id=8dcf0ff8-b97b-41be-b761-2cd409014d8a ip=127.0.0.1
<Response streamed [401 UNAUTHORIZED]>
.2020-06-10 07:45:52,536 alerta.app[8385]: [WARNING] LDAP LOGIN request_id=78860ae7-4a27-4747-a4b3-fb4d61fe43e5 ip=127.0.0.1
<Response streamed [401 UNAUTHORIZED]>
.
```

alerta/auth/basic_ldap.py
satterly: I've tried a few things and nothing works so your approach is fine. Can you rename `main_auth.py` to `login.py`. That makes it consistent with `logout.py`.

Also just wondering why you moved `def login()` in `basic.py` to the bottom of the file? If it makes no difference can you move it back? These sorts of changes create lots of unnecessary noise in a PR making it harder to judge what's changed.

alerta/auth/__init__.py
```diff
@@ -20,12 +20,13 @@ def init_auth(app):
     if app.config['AUTH_PROVIDER'] == 'ldap':
         try:
             import ldap  # noqa
-            from . import basic_ldap  # noqa
         except ImportError:
             raise RuntimeError('Must install python-ldap to use LDAP authentication module')
     else:
         from . import basic  # noqa
 
+    from . import main_auth  # noqa
```
satterly, line 28: Delete this line. And line 38 becomes ...
```
from . import github, login, logout, oidc, userinfo   # noqa isort:skip
```

alerta/auth/basic.py
```diff
@@ -63,49 +63,6 @@ def signup():
     return jsonify(token=token.tokenize)
 
 
-@auth.route('/auth/login', methods=['OPTIONS', 'POST'])
```
satterly, line 66: Move this back to where to here.

alerta/auth/basic.py
```diff
@@ -169,3 +126,44 @@ def reset(hash):
         return jsonify(status='ok', message='password reset successful')
     else:
         raise ApiError('invalid password reset hash', 400)
+
+
+def login():
```
satterly, line 131: Revert this move.

alerta/auth/main_auth.py
```diff
@@ -0,0 +1,15 @@
+from flask import current_app
```
satterly, line 1: Change this filename to `login.py`.

alerta/views/alerts.py
```diff
@@ -256,6 +256,8 @@ def update_attributes(alert_id):
         raise ApiError('failed to update attributes', 500)
 
 # delete
+
+
```
satterly, line 260: Revert this whitespace change.

requirements-dev.txt
```diff
@@ -4,6 +4,7 @@ pylint==2.4.4
 pytest
 pytest-cov
 python-dotenv
+python-ldap
```
satterly, line 7: Remove.

satterly:
@frasy Any chance you can look at this? If not, I'll close it for now. Thanks.

frasy:
> @frasy Any chance you can look at this? If not, I'll close it for now. Thanks.

Hi, I've been out a few days, I will review this and I will push a PR today or tomorrow at last

satterly:
I suggest the first thing you should do is rebase with master and the resolve conflicts. Then make the suggested changes and hopefully the tests should pass. Then I'll review the actual LDAP code changes to see if we need to make any improvements there. Thanks.

requirements-dev.txt
frasy: I added this here because Travis CI failed when I did the last code changes. Anyway, I will remove it and I will see if Travis CI still fails of if this is no longer needed

frasy:
Hi @satterly, I have already done the proposed changes
To have Travis-CI working I had to keep the python-ldap module in the requirements-dev.txt file

I will work in a LDIF file to provide in order you can do the tests.

KlavsKlavsen:
Whats the status of this? it seems done, no ?

KlavsKlavsen:
This would allow me to filter on MemberOf on logins - so I could ensure login is only allowed to users who are a member of a certain group - which I can't any other way to do (when using AD/LDAP groups and AD/LDAP login)

frasy:
Hi @KlavsKlavsen, this should enable the LDAP filtering. Waiting for @satterly to come back from holidays to retake the work on this

satterly:
I've had another look at this and it looks good. My only concern is that the configuration seems to be getting more and more complex. This is due to the `DOMAINS` lookup settings that means everything needs to be a dict even if there's only one value.

How many people actually use the multiple domains/basedn feature possible with `LDAP_DOMAINS`? If it isn't being used then it would make things much simpler if removed.

Thoughts? @frasy @KlavsKlavsen @Anderen2 @PMDubuc @carlospeon @raffis @olivluca

KlavsKlavsen:
I think one should have a look at f.ex. how LDAP works for grafana: https://grafana.com/docs/grafana/latest/auth/ldap/ - it really wasn't that hard to get that working with everything I needed (including defining which AD groups that render access and at which level).

satterly:
@KlavsKlavsen Yes, I have looked a Grafan, Okta, Auth0 and others. None of them support this concept of multiple domains/dns. So I guess you're suggesting I can remove the `DOMAINS` mapping which just adds an additional level of complexity.

KlavsKlavsen:
I'd say that when all those do not support it - its really not needed.
If one HAS multiple AD domains (or LDAP for that matter) - you usually join the tree -and can then just point on those LDAP servers that sees the entire tree - with all domains.

satterly (approved):

satterly:
Given the feedback, I'm going to merge this PR and then work on removing the `DOMAINS` code in another PR.

PR #1218, opened 2020-05-05
added LDAP_DEFAULT_DOMAIN to allow ldap login with bare username

olivluca:
In my setup (and I think it's the same in most places) ldap login is done with the bare username and not with the email address. This pull request adds a LDAP_DEFAULT_DOMAIN option so, if that is specified, it allows to login with just an username. If it's not set it will behave as before.


KlavsKlavsen:
Please merge this - its definetely useful

Issue #1091, opened 2019-10-02
Grafana Webhook Integration:  Looks for `tags` in the wrong part of the json payload


travis-bear:
**Issue Summary**
Grafana webhook doesn't convert the tags sent by Grafana into custom attributes.  This appears to be because in Grafana.py, it's looking in the wrong part of the incoming json document for the tags.

Here's the incoming alert payload from Grafana version 6.3.6, which is the newest version:

    { 
       "evalMatches":[ 
          { 
             "value":323.4,
             "metric":"samples",
             "tags":{ 
             }
          }
       ],
       "message":"Here's my message",
       "ruleId":12,
       "ruleName":"Samples Appended alert",
       "ruleUrl":"http://grafana.monitor.aort.<...>.com/d/itBdm3dZz/prometheus-health?fullscreen\\u0026edit\\u0026tab=alert\\u0026panelId=3\\u0026orgId=1",
       "state":"alerting",
       "tags":{ 
          "arbitrary_key":"arbitrary_val",
          "dashboard":"my dashboard",
          "slack":"my_slack",
          "tsg":"my tsg"
       },
       "title":"[Alerting] Samples Appended alert"
    }

Here's the Grafana alert config:

<img width="683" alt="Screen Shot 2019-10-02 at 4 04 09 PM" src="https://user-images.githubusercontent.com/505972/66089546-b63caa80-e534-11e9-9e58-f3959a20a76c.png">

In Alerta's grafana code (https://github.com/alerta/alerta/blob/master/alerta/webhooks/grafana.py#L33) it appears to be building custom attributes from the tag data from some `match` stanza rather than the top-level `tags` stanza where the sent tags are actually located.

satterly:
This must have changed at some point because the test cases are based on payloads that look like this ...
```
        {
            "evalMatches": [
                {
                    "value": 97.007606,
                    "metric": "user",
                    "tags": {
                        "__name__": "netdata_cpu_cpu_percentage_average",
                        "chart": "cpu.cpu0",
                        "dimension": "user",
                        "family": "utilization",
                        "instance": "zeta.domain",
                        "job": "monitoring",
                        "info.host_id": "i-0d0721c7f97545d43"
                    }
                }
            ],
            "message": "boom!",
            "ruleId": 7,
            "ruleName": "Zeta CPU alert",
            "ruleUrl": "https://grafana.domain.tld/dashboard/db/alarms?fullscreen&edit&tab=alert&panelId=1&orgId=1",
[...]

PR #1093, opened 2019-10-03
Add support for alert rule tags to Grafana webhook

satterly:
Fixes #1091

PR #1234, opened 2020-06-04
webhooks/grafana: Add support for Grafana tags

psyhomb:
Fix for the issue https://github.com/alerta/alerta/issues/1091.


satterly:
Hey thanks for the PR but I already had a go at this #1093 but the tests were failing. So I started again, fixed and expanded the tests and added support for `customer` override as well in #1243 and #1245 so I'm going to close this PR. Thanks again for your contribution.

PR #1265, opened 2020-07-02
Fix server-side sort for custom attributes

### LINKED ISSUE #1262: Can't sort custom columns since v8+
**Issue Summary**
Since v8+ release (bug still present in 8.0.2), there is an error when trying to sort by custom column with this error message :

```
Missing FROM-clause entry for table "attributes" LINE 7: ORDER BY attributes.clusterType DESC ^ (500)
```

**Environment**
- OS: Linux
- API version: 8.0.2
- Deployment: Kubernetes (Docker based, )

- Database: Postgres

- Server config:
  Auth enabled? Yes
  Auth provider? Basic

- web UI version: 8.0.0
- CLI version: 8.0.2

**To Reproduce**
- Create custom columns that are mapped on attributes
- Try to sort by any custom column name and you'll get the following error message while your column not being sorted as expected.

```
Missing FROM-clause entry for table "attributes" LINE 7: ORDER BY attributes.clusterType DESC ^ (500)
```

Note : sorting other columns work as expected, only custom columns are faulty

We redeployed many times from scratch, burning the database and have a clean install, the problem is still there.

satterly:
Fixes #1262

### DIFF
--- a/alerta/database/backends/postgres/utils.py
+++ b/alerta/database/backends/postgres/utils.py
@@ -78,6 +78,8 @@ class QueryBuilderImpl(QueryBuilder):
                     sort.append('s.code ' + direction)
                 elif sort_by == 'status':
                     sort.append('st.state ' + direction)
+                elif sort_by.startswith('attributes'):
+                    sort.append("attributes->'{}' {}".format(sort_by.replace('attributes.', ''), direction))
                 elif sort_by == 'createTime':
                     sort.append('create_time ' + reverse_sort(direction))
                 elif sort_by == 'receiveTime':


robinlem:
Hi satterly, your latest bluid on dockerhub  doesn't seams have your fix. Have you planned to build it soon ?
Regards

fredma:
Hi there, just a comment to say that this is no longer working on the latest version.

WEBUI 8.5.0 
Alerta API 8.5.0

This is because the sorting code has a static list of `Alerts.VALID_PARAMS` with `attributes` being the valid param.
With a custom column configured, I currently have the exception `Sorting by 'attributes.jira' field not supported. `

Maybe allowing the VALID_PARAMS to use the COLUMNS list in the Alerta configuration file could work?

Issue #1315, opened 2020-09-25
New Relic built-in webhook seems broken in alerta 8.0.2


pantelis-karamolegkos:
**Issue Summary**

When creating a web hook in New Relic pointing to `http://my-alerta/api/webhooks/newrelic` and sending a test notification, I get __BAD REQUEST/400__

**Environment**
- OS: N/A
- API version: `alerta/alerta-web:8.0.2`
- Deployment: `docker-compose`
- For self-hosted, WSGI environment: N/A

- Database: Postgres

- Server config:
  Auth enabled? No
  Auth provider? N/A
  Customer views? No
  (or provide the `/config` endpoint output)

- web UI version: `alerta/alerta-web:8.0.2`
- CLI version: `alerta/alerta-web:8.0.2`

**To Reproduce**

 
Spin up a `docker-compose` stack as follows:

```
version: '3.1'
services:
  web:
    # build:
    #   context: .
    #   args:
    #     ALERTA_BASE_VERSION: ${ALERTA_BASE_VERSION}
    image: alerta/alerta-web:8.0.2
    volumes:
      - ./alertad.conf:/app/alertad.conf
    ports:
      - 8080:8080
    depends_on:
      - db
    environment:
      - DEBUG=5  # remove this line to turn DEBUG off
      - DATABASE_URL=postgres://postgres:postgres@db:5432/monitoring
    restart: always
  db:
    image: postgres
    volumes:
      - ./pg-data:/var/lib/postgresql/data
    environment:
      POSTGRES_DB: monitoring
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: postgres
    restart: always
```

Expose it via `ngrok`

Create webhook type notification channel in `New Relic` and provide as base url: `http://my-ngrok-url/api/webooks/newrelic`

Send a test notification to the above notification channel

**Expected behavior**
Alert should be registered normally

**Screenshots**

**Additional context**

Logs:

```
2020-09-25 11:23:30,334 DEBG 'uwsgi' stdout output:
2020-09-25 11:23:30,334 alerta.app[24]: [INFO] Webhook Request:
POST http://64444444444e.ngrok.io/api/webhooks/newrelic
Content-Type: application/json
Content-Length: 1021
Accept: text/plain, application/json, application/*+json, */*
Charset: utf-8
Tracestate: 1@nr=0-1.846443-1601033007894
Traceparent: 00-6a44a38ec07b5fbede-01
[...]

PR #1318, opened 2020-09-26
Fixing newrelic webhook

pkaramol:
The test notification of New Relic was broken.

I managed to identify 3 issues (perhaps there are more, but for the time being the test notification seems to be processed)

a) The control statements setting the value of `severity` ended up with `severity="info"` which is not an acceptable value (although this can be altered in the config)

b) the event ended up as empty value

c) the resource also ended up as empty value

I also simplified the attributes not to include HTML tags but this can be changed (at some point I though the HTML was breaking the process)

I had to learn (the very hard way) that the `400` was caused by the `Alert` object not being returned appropriately.

So my question is why there were no logs to point to the right direction (unless I missed something)

I had this in my local `docker-compose` stack set

```
    environment:
      - DEBUG=True  # remove this line to turn DEBUG off
```

I hope it addresses #1315

### DIFF
--- a/alerta/webhooks/newrelic.py
+++ b/alerta/webhooks/newrelic.py
@@ -1,13 +1,15 @@
 from typing import Any, Dict
 
 from alerta.models.alert import Alert
-
+from alerta.models.alarms.alerta import SEVERITY_MAP
 from . import WebhookBase
 
 JSON = Dict[str, Any]
+UNDETERMINED ="Untetermined"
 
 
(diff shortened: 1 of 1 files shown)


satterly:
Can you paste the New Relic test notification into this PR and I'll include it in the tests? Thanks.

satterly:
Don't worry, I found it in the origin Issue you raised. 👍

Issue #1208, opened 2020-04-23
Check LDAP auth works with Active Directory LDAPS


satterly:
**Is your feature request related to a problem? Please describe.**

Does Alerta support SLDAP since Microsoft is deprecating normal LDAP? @cisco92

**Describe the solution you'd like**

Alerta LDAP auth should continue to work.

**Describe alternatives you've considered**

None.

**Additional context**

https://support.microsoft.com/en-us/help/4520412/2020-ldap-channel-binding-and-ldap-signing-requirements-for-windows
https://kb.sos-berlin.com/pages/viewpage.action?pageId=18778435

See also https://github.com/python-ldap/python-ldap/issues/309

wagnst:
I think it is actually not working...

```
2020-04-29 14:46:42,974 DEBG 'uwsgi' stdout output:
*** <ldap.ldapobject.SimpleLDAPObject object at 0x7f5ea44c2410> ldaps://ldap.xxx.de:636/ - SimpleLDAPObject.set_option
((17, 3), {})
=> result:
None

2020-04-29 14:46:42,975 DEBG 'uwsgi' stdout output:
*** <ldap.ldapobject.SimpleLDAPObject object at 0x7f5ea44c2410> ldaps://ldap.xxx.de:636/ - SimpleLDAPObject.simple_bind
(('(&(objectClass=person)(uid=admin))', 'xxx', None, None), {})

2020-04-29 14:46:43,081 DEBG 'uwsgi' stdout output:
=> LDAPError - SERVER_DOWN: {'desc': "Can't contact LDAP server", 'info': '(unknown error code)'}

2020-04-29 14:46:43,086 DEBG 'uwsgi' stdout output:
2020-04-29 14:46:43,081 flask.app[25]: [ERROR] {'desc': "Can't contact LDAP server", 'info': '(unknown error code)'} request_id=4a743f2199d4b9ad0011d82ede01e8c1 ip=10.244.0.32
Traceback (most recent call last):
  File "/venv/lib/python3.7/site-packages/alerta/auth/basic_ldap.py", line 52, in login
    ldap_connection.simple_bind_s(userdn, password)
  File "/venv/lib/python3.7/site-packages/ldap/ldapobject.py", line 445, in simple_bind_s
    msgid = self.simple_bind(who,cred,serverctrls,clientctrls)
  File "/venv/lib/python3.7/site-packages/ldap/ldapobject.py", line 439, in simple_bind
[...]

PR #1323, opened 2020-09-29
implement PR1219 with requested changes and naming adjustment from me.

KlavsKlavsen:


KlavsKlavsen:
Resolves #1208

KlavsKlavsen:
giving CA cert is also necessary as Microsoft AD normally runs with its own CA - and uses that for LDAP service - which often is NOT part of the system CA list that Alerta would otherwise be using - and in the case of running it via docker - its the only way, to avoid having to modify the docker image to include the CACert - and allowing self-signed certs is essentially disabling the security :(

satterly:
👍  Thanks. I will fix the linting error.

PR #1329, opened 2020-10-09
Fix CloudWatch webhook bug

psyhomb:
If `Dimensions` list is empty alert is never going to be shown on the Alerta dashboard and this is definitely not a desired behaviour.
By making this change if `Dimensions` list is empty value for `resource` variable will be set to topic name and alert will be shown on the Alerta dashboard.


satterly:
Thanks 👍

PR #1339, opened 2020-10-20
[Enhancement] Ability to use tags in Pingdom to set 'environment' and 'group' of alert

wikro:
To change `environment` of alert triggered by a Pingdom webhook add the tag `environment_YourEnvironment` to the check, and `environment` gets set to `yourenvironment` (Pingdom changes uppercase to lowercase).

To change `group` of the alert triggered by a Pingdom webhook add the tag `group_YourGroupName` to the check, and `group` gets set to `yourgroupname` (Pingdom changes uppercase to lowercase).

### DIFF
--- a/alerta/webhooks/pingdom.py
+++ b/alerta/webhooks/pingdom.py
@@ -29,7 +29,7 @@ class PingdomWebhook(WebhookBase):
             severity = alarm_model.DEFAULT_NORMAL_SEVERITY
 
         if len(payload['tags']) > 0:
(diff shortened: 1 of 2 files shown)


satterly:
Thanks for the PR. Can you update the webhook test for Pingdom to include `tags` in the [response body](https://github.com/alerta/alerta/blob/master/tests/test_webhooks.py#L371) and then [check for them](https://github.com/alerta/alerta/blob/master/tests/test_webhooks.py#L821-L825) in the alert?

wikro:
Closing the PR since this may probably only be something for myself, and a custom webhook was easy to implement instead.

PR #1344, opened 2020-10-23
Use guest scope as default

escoand:
Currently the default scope of a new user is `user`.

This is not ideal when using `oidc` (and maybe also others) as auth because the user gets created with first login attempt. So the user has maybe more rights than desired.

With this change the default scope is changed to `guest`.

### DIFF
--- a/alerta/database/backends/mongodb/base.py
+++ b/alerta/database/backends/mongodb/base.py
@@ -1498,7 +1498,7 @@ class Backend(Database):
             response = self.get_db().perms.find_one({'match': match}, projection={'scopes': 1, '_id': 0})
             if response:
                 scopes.extend(response['scopes'])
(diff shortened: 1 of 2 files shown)


satterly:
Why not just set `USER_DEFAULT_SCOPES` to `read` instead of `read` and `write`? 

```
USER_DEFAULT_SCOPES = ['read']
```

satterly:
See also #1346

escoand:
> Why not just set `USER_DEFAULT_SCOPES` to `read` instead of `read` and `write`? 
> 
> ```
> USER_DEFAULT_SCOPES = ['read']
> ```

I need all three builtin roles. So setting the default user scope to read forces me to create an additional user defined role for r/w. But I'm also fine with your PR.

satterly:
Closing in favour of #1346

PR #1148, opened 2020-02-13
Support user-defined list of admin roles

satterly:
Two options available ..
*defaults*
```
DEFAULT_ADMIN_ROLE = 'admin'
ADMIN_ROLES = [DEFAULT_ADMIN_ROLE]
```
*example*
```
DEFAULT_ADMIN_ROLE = 'ops'
ADMIN_ROLES = ['ops', 'devops', 'coolkids']
```

**Note: the `DEFAULT_ADMIN_ROLE` must be contained within the list of `ADMIN_ROLES` or unpredictable things might happen.**

Fixes #1146

PR #1346, opened 2020-10-24
Support user-defined list of user roles

satterly:
Add config option to allow Alerta admins to set the default roles for users when they log in for the first time.

**Defaults**
```
DEFAULT_USER_ROLE = 'user'
USER_ROLES = [DEFAULT_USER_ROLE]
```
For example, to set default user role to "very-limited-role" ...
```
DEFAULT_USER_ROLE = 'very-limited-role'
USER_ROLES = [DEFAULT_USER_ROLE]
```
or to nothing ...
```
USER_ROLES = []
```

See also #1148

### DIFF
--- a/alerta/auth/basic.py
+++ b/alerta/auth/basic.py
@@ -23,7 +23,7 @@ def signup():
         raise ApiError(str(e), 400)
 
     # set sign-up defaults
-    user.roles = ['user']
+    user.roles = current_app.config['USER_ROLES']
     user.email_verified = False
 
     # check allowed domain
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -71,7 +71,7 @@ def login():
     user = User.find_by_username(username=login)
     if not user:
         user = User(name=username, login=login, password='', email=email,
-                    roles=[], text='LDAP user', email_verified=email_verified)
+                    roles=current_app.config['USER_ROLES'], text='LDAP user', email_verified=email_verified)
         try:
             user = user.create()
         except Exception as e:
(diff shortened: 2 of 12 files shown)


escoand:
Do I understand it right, we now have still the 3 roles but newly created users aren't part of any of them?

satterly:
Why do you think that?

escoand:
Because of
```diff
-        return sorted(set(scopes)) or current_app.config['USER_DEFAULT_SCOPES']
+        return sorted(set(scopes))
```

escoand:
Anyway, my question was how to implement my initial request from https://github.com/alerta/alerta/pull/1344 with this new configuration? Do I have to set `USER_ROLES=["guest"]`? The purpose of these new options are not obvious to me.

satterly:
So you should use `USER_ROLES=['guest']`

satterly:
> Because of
> 
> ```diff
> -        return sorted(set(scopes)) or current_app.config['USER_DEFAULT_SCOPES']
> +        return sorted(set(scopes))
> ```

User roles are assigned when user is created now.

```
>    - user.roles = ['user']
>    + user.roles = current_app.config['USER_ROLES']
```

escoand:
Sorry, but I still don't get it. I'm using this configuration:

```python
USER_ROLES = ["guest"]
USER_DEFAULT_SCOPES = ["write:alerts", "read:heartbeats", "read:blackouts", "read:customers"]
GUEST_DEFAULT_SCOPES = ["read:alerts", "read:customers"]
```

The new user correctly gets the `guest` role but with the permissions of `USER_DEFAULT_SCOPES`. Additionally the `user` role has no permissions anymore. So it seems I've renamed the user role to guest and overwritten the built in guest role.

PR #1351, opened 2020-10-29
Refactor LDAP auth to simplify config

### LINKED ISSUE #951: change the basedn formatter to have flexibility in query ldap

satterly:
All supported LDAP configuration settings...
```
# BasicAuth using LDAP
LDAP_URL = ''  # eg. ldap://localhost:389
LDAP_BASEDN = ''
LDAP_CACERT = ''  # Path to CA certificate to verify LDAPS connection against
LDAP_ALLOW_SELF_SIGNED_CERT = False
LDAP_DOMAINS = {
    # 'planetexpress.com': 'cn=%s,ou=people,dc=planetexpress,dc=com',
    # 'sicredi.com': 'uid={0},cn={0:.1},cn=users,dc=sicredi,dc=com,dc=br'
}
LDAP_BIND_USERNAME = ''  # required if using LDAP_SEARCH_QUERY eg. uid=admin,ou=users,dc=domain,dc=com
LDAP_BIND_PASSWORD = ''  # required if using LDAP_BIND_USERNAME
LDAP_USER_BASEDN = ''  # BASEDN for user search (default: LDAP_BASEDN)
LDAP_USER_FILTER = ''  # eg. (cn={username})
LDAP_USER_NAME_ATTR = 'cn'  # eg. cn or displayName
LDAP_USER_EMAIL_ATTR = 'mail'  # eg. mail or email
LDAP_GROUP_BASEDN = ''  # BASEDN for group search (default: LDAP_BASEDN)
LDAP_GROUP_FILTER = ''  # eg. (&(member={userdn})(objectClass=group))
LDAP_GROUP_NAME_ATTR = 'memberOf'  # eg. memberOf or cn
LDAP_DEFAULT_DOMAIN = ''  # if set allows users to login with bare username
ALLOWED_EMAIL_DOMAINS = []
```

Example...
```
AUTH_PROVIDER='ldap'
LDAP_URL='ldap://localhost:389'
LDAP_BASEDN='ou=people,dc=planetexpress,dc=com'
LDAP_BIND_USERNAME='cn=admin,dc=planetexpress,dc=com'
LDAP_BIND_PASSWORD='GoodNewsEveryone'
LDAP_USER_FILTER='(&(uid={username})(objectClass=inetOrgPerson))'
LDAP_USER_NAME_ATTR='cn'
LDAP_USER_EMAIL_ATTR='mail',
LDAP_GROUP_FILTER='(&(member={userdn})(objectClass=group))'
LDAP_GROUP_NAME_ATTR='cn'
ALLOWED_EMAIL_DOMAINS=['planetexpress.com']
````

Improves #1206 Fixes #951

/cc @frasy @KlavsKlavsen @Anderen2 @PMDubuc @carlospeon @raffis @olivluca

### DIFF
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -1,5 +1,3 @@
-import sys
-
 import ldap  # pylint: disable=import-error
 from flask import current_app, jsonify, request
 
@@ -11,17 +9,18 @@ from alerta.utils.audit import auth_audit_trail
 
 
 def login():
-    # Allow LDAP server to use a self signed certificate
-    if current_app.config['LDAP_ALLOW_SELF_SIGNED_CERT']:
-        ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_ALLOW)
 
+    # LDAP certificate settings
     if current_app.config['LDAP_CACERT']:
         ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_HARD)
         ldap.set_option(ldap.OPT_X_TLS_CACERTFILE, current_app.config['LDAP_CACERT'])
 
-    # Retrieve required fields from client request
+    # Allow LDAP server to use a self-signed certificate
+    if current_app.config['LDAP_ALLOW_SELF_SIGNED_CERT']:
+        ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_ALLOW)
+
     try:
-        login = request.json.get('username', None) or request.json['email']
+        login = request.json.get('username') or request.json['email']
         password = request.json['password']
     except KeyError:
         raise ApiError("must supply 'username' and 'password'", 401)
@@ -32,32 +31,29 @@ def login():
     try:
         if '\\' in login:
             domain, username = login.split('\\')
-            email = ''
-            email_verified = False
         else:
             username, domain = login.split('@')
-            email = login
-            email_verified = True
     except ValueError:
         if current_app.config['LDAP_DEFAULT_DOMAIN']:
             username = login
             domain = current_app.config['LDAP_DEFAULT_DOMAIN']
-            email = username + '@' + domain
-            email_verified = True
         else:
             raise ApiError('expected username with domain', 401)
 
     # Validate LDAP domain
-    if domain not in current_app.config['LDAP_DOMAINS'] and \
-       domain not in current_app.config['LDAP_DOMAINS_SEARCH_QUERY']:
+    if (domain not in current_app.config['ALLOWED_EMAIL_DOMAINS']
+            and domain not in current_app.config['LDAP_DOMAINS']):
         raise ApiError('unauthorized domain', 403)
 
     # Initialise ldap connection
     try:
-        trace_level = 2 if current_app.debug else 0
+        trace_level = 2 if current_app.debug else 0  # XXX - do not set in production environments
         ldap_connection = ldap.initialize(current_app.config['LDAP_URL'], trace_level=trace_level)
     except Exception as e:
         raise ApiError(str(e), 500)
+
+    # Set default base DN for user and group search
+    base_dn = current_app.config['LDAP_BASEDN']
 
     # If user search filter exist
     #   Search the user using the provided User Search filter for the current domain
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: alerta/auth/basic_ldap.py [hunk 1]
```diff
@@ -68,82 +64,83 @@ def login():
...
-    domain_search_query = current_app.config.get('LDAP_DOMAINS_SEARCH_QUERY', {})
-    base_dns = current_app.config.get('LDAP_DOMAINS_BASEDN', {})
-    user_base_dn = current_app.config.get('LDAP_DOMAINS_USER_BASEDN', {})
-    if domain in domain_search_query:
-        ldap_bind_username = current_app.config.get('LDAP_BIND_USERNAME', '')
-        ldap_bind_password = current_app.config.get('LDAP_BIND_PASSWORD', '')
+    user_filter = current_app.config['LDAP_USER_FILTER']
+    user_base_dn = current_app.config['LDAP_USER_BASEDN']
+    if user_filter:
+        ldap_bind_username = current_app.config['LDAP_BIND_USERNAME']
```
KlavsKlavsen, line 70: what happens if LDAP_BIND_USERNAME isn't set? no friendly error message hinting at whats up?

KlavsKlavsen, line 70: I'm assuming current_app.config doesn't fail if that entry isn't in the array ?

KlavsKlavsen: if it does - it should probably be moved into the try :)

KlavsKlavsen:
I'll test it out monday - it looks good :)

alerta/auth/basic_ldap.py
frasy: As far as I can see this cannot fail because it is already part of ```alerta/settings.py``` file, so the property will always exist although it will be blank string if it is not added in the user config file

satterly: @frasy is right. As long a default value already exists in `settings.py` it won't throw a key error.

frasy:
I see it everything ok

satterly:
> I'll test it out monday - it looks good :)

@KlavsKlavsen I'm going to merge this PR as is because it's blocking some other work I need to do but please still do test and if you find any problems let me know and I'll address them in a new PR. Thanks.

KlavsKlavsen:
How is LDAP_GROUP_FILTER used to ensure user MUST be a member of a specific group, to be allowed to login? the example does not show that unfortunately :(

KlavsKlavsen:
I'm trying with: (&(objectClass=group)(cn=$groupname)) - to see if that doesn't work.
I assume ADMIN_ROLES setting is not used anymore?

KlavsKlavsen:
as I read the code.. it searches for groups matching the group filter - and IF the filter (using either username, email or userdn) matches groups - those will be added to the list of groups..
so one would need to add the group name from AD (the cn=$groupname) part to list of ADMIN_ROLES
and ensure there is no groups in different parts of the tree - with same cn?

satterly:
Can we do this on Slack? It's too difficult otherwise. https://join.slack.com/t/alerta/shared_invite/zt-ipdorryu-e2rTME2jxIbbBrYH5LvWqQ Or Google Hangouts?

PR #1392, opened 2020-11-25
Add take_note plugin method

### LINKED ISSUE #1382: [Feature request] Allow processing notes in plugins

I'd like to add some logic when users add notes, for example populate alert attributes based on the note content, triggering escalation, ...
Notes added with an action can be treated with a plugin using `take_action`, but notes added without an action cannot as they're going through a different api route: `note`.

Would it make sense to add another plugin method to hook custom logic on notes ?
Or could notes be fed into the existing `take_action` plugin method with an action name like `_note` ?

I've considered using a dedicated custom action instead of relying on vanilla notes, but the UI does not offers custom actions with text (although I think hitting `/action` with `{"action":"customAction","text":"foo"}` should work).

**Additional context**

I can submit a PR if you confirm you're interested in the feature and want it implemented in the backend. If you prefer to add custom actions with text in the UI it would also totally solve my issue but I likely won't be able to help you implement it.

hugoShaka:
Closes #1382 

This PR:
* adds the `take_note()` method for plugins
* processes notes when created or updated with a text
* adds minimal tests cases: trigger plugin on create, trigger plugin on update, check attributes changes are properly saved

I mimicked the `take_action` support for the plugin returning either `Alert` or `Tupple(Alert, something)`, this looks like backward compatibility for legacy plugins, I'm not sure if keeping it makes sense.

### DIFF
--- a/alerta/views/alerts.py
+++ b/alerta/views/alerts.py
@@ -13,7 +13,7 @@ from alerta.models.enums import Scope
 from alerta.models.metrics import Timer, timer
 from alerta.models.switch import Switch
 from alerta.utils.api import (assign_customer, process_action, process_alert,
-                              process_delete, process_status)
+                              process_delete, process_note, process_status)
 from alerta.utils.audit import write_audit_trail
 from alerta.utils.paging import Page
 from alerta.utils.response import absolute_url, jsonp
@@ -587,6 +587,7 @@ def add_note(alert_id):
     if not alert:
         raise ApiError('not found', 404)
 
+    alert, note_text = process_note(alert, note_text)
     note = alert.add_note(note_text)
 
     write_audit_trail.send(current_app._get_current_object(), event='alert-note-added', message='', user=g.login,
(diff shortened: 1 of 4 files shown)


hugoShaka:
Lint and tests are passing but the Slack action is failing, it's likely due to the var `SLACK_WEBHOOK_URL` being unavailable on jobs ran by someone outside of the project.

alerta/views/alerts.py
```diff
@@ -649,6 +650,9 @@ def update_note(alert_id, note_id):
 
     update = request.json
     update['user'] = g.login
+    if update.get('text', ''):
+        alert, note_text = process_note(alert, update['text'])
+        update['text'] = note_text
```
satterly, line 655: Can you move the above 3 lines to below the `write_audit_trail.send()`? And adjust to look like ...

```
    _, update['text'] = process_note(alert, update['text'])
    updated = note.update(**update)
    if updated:
        return jsonify(status='ok', note=updated.serialize)
    else:
        raise ApiError('failed to update note', 500)
```
I think it should run the hook whether there is text or not. Someone might want to add text if missing, for example.

satterly (changes requested):
Excellent work. One minor suggestion. 👍

alerta/views/alerts.py
hugoShaka: Fix done (I also changed function signature as `None` can be passed) , if you're ok I can rebase and squash

satterly: Nice. Thanks! No need to rebase/squash. I'll merge as is.

satterly (approved):
LGTM 👍

Issue #1391, opened 2020-11-25
LDAP authentication to Active Directory fails when LDAP_USER_FILTER is set - Always


PedroMSantosD:
**Issue Summary**
Authentication to Active Directory using LDAP always fails when LDAP_USER_FILTER is set. Any user filter passed to the ldap library returns a list of 5 lines or more;  one (or more) for the actual User filter match, and the remaining ones for Active Directory partitions, which are not expected by the basic_ldap.py code.

Example:
alerta.conf
```
LDAP_USER_BASEDN = 'DC=mydomain,DC=com'
LDAP_USER_NAME_ATTR = 'sAMAccountName'
LDAP_USER_EMAIL_ATTR = 'mail'
LDAP_USER_FILTER = '(&(objectCategory=Person)(objectCategory=Person)(!(UserAccountControl:1.2.840.113556.1.4.803:=2))(sAMAccountName={username}))'
```
Yields [response](https://github.com/alerta/alerta/blob/master/alerta/auth/basic_ldap.py#L83)

```
[('CN=Pedro '
  'Santos,OU=Foo,OU=Bar,DC=mydomain,DC=com',
  {'mail': [b'pesantos@mydomain.com'], 'sAMAccountName': [b'pesantos']}),
 (None,
  ['ldap://mydomain.com/CN=Configuration,DC=mydomain,DC=com']),
 (None,
  ['ldap://DomainDnsZones.mydomain.com/DC=DomainDnsZones,DC=mydomain,DC=com']),
 (None,
  ['ldap://ForestDnsZones.mydomain.com/DC=ForestDnsZones,DC=mydomain,DC=com']),
 (None,
  ['ldap://mydomain.com/CN=Schema,CN=Configuration,DC=mydomain,DC=com'])],
```

Since this list has 5 elements (one for each ADDS partition), code [Fails](https://github.com/alerta/alerta/blob/master/alerta/auth/basic_ldap.py#L91) with invalid search query for domain

**Environment**
- OS: 
```
# cat /etc/centos-release
CentOS Linux release 7.8.2003 (Core)
# uname -a
Linux monitoring-alerta-server-1.internal 3.10.0-1127.19.1.el7.x86_64 #1 SMP Tue Aug 25 17:23:54 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

```
- API version:
```
config# pip freeze
alerta==8.0.0
alerta-icinga2api==1.0.1
alerta-mattermost==1.4
[...]

PR #1394, opened 2020-11-26
removing None tupples from Active directory partitions on LDAP respon…

PedroMSantosD:
code to fix https://github.com/alerta/alerta/issues/1391.

(1) Handles replies from all Active Directory Partitions, removing None tupples before failing.
(2) adds timeout to Ldap search if parameter LDAP_QUERY_TIMEOUT_SECONDS is added to configuration file.

### DIFF
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -1,5 +1,4 @@
 import ldap  # pylint: disable=import-error
-import logging
 from flask import current_app, jsonify, request
 
 from alerta.auth.utils import create_token, get_customers, not_authorized
--- a/alerta/auth/basic_ldap.py
+++ b/alerta/auth/basic_ldap.py
@@ -1,4 +1,5 @@
 import ldap  # pylint: disable=import-error
+import logging
 from flask import current_app, jsonify, request
 
 from alerta.auth.utils import create_token, get_customers, not_authorized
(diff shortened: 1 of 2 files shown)


alerta/auth/basic_ldap.py
```diff
@@ -88,10 +93,21 @@ def login():
...
+            for result_tupple in result:
+                if None not in result_tupple:
+                    tmp_result.append(result_tupple)
+            logging.debug('ADDS - Filtered LDAP response has {} entries'.format(len(tmp_result)))
+            if len(tmp_result) > 1:
+                raise ApiError('invalid search query for domain "{}"'.format(domain), 500)
+            elif len(tmp_result) == 0:
+                raise ApiError('invalid username or password', 401)
+            else:
+                result = tmp_result
```
satterly, line 108: The above can be removed by just modifying lines 82-87 to use a list comprehension ...
```
    if user_filter:
        result = [r for r in ldap_connection.search_s(
            base=user_base_dn or base_dn,
            scope=ldap.SCOPE_SUBTREE,
            filterstr=user_filter.format(username=username),
            attrlist=user_attrs
        ) if None not in r]
```

PedroMSantosD: Thanks for the suggestion. I have implemented your changes and added the optional variable to the test file so they complete.

satterly: 👍

satterly:
LGTM ,Thanks!

PR #1438, opened 2021-01-06
fix root dictconfig key in logging.py

wo250012:
Submitting a fix for the root key in the dictconfig. According to https://docs.python.org/2/library/logging.config.html#dictionary-schema-details a root config should be it's own key outside of loggers.


satterly:
Thanks. 👍  (btw, the failing integration has nothing to do with your change)

PR #1442, opened 2021-01-07
request filter fixes in logging.py

3IMOH:
Change order of filters to add request parameters to logging entry. Changed request filter to ignore non-http request log entries.


3IMOH:
hmm...just noticed my last fix for the root key is included in this pull req. Not sure why as I updated my repo before this most recent change. sorry about that.

PR #1446, opened 2021-01-11
refactor: create "query builders" for all data models

satterly:
Required for server-side query filtering. See alerta/alerta-webui#448


MaiHistoryCompanyLimited:
alerta/alerta-webui#448

PR #1482, opened 2021-03-12
feat(auth): support multiple auth providers

### LINKED ISSUE #1389: Authentification Basic + Saml2
Hello,

How enable authentification basic and saml2 ? 
The variable AUTH_PROVIDER is "basic" or "saml2".. but i wish "basic" AND "saml2"

Thx

satterly:
Fixes #1389


satterly:
Too complicated. Just need to add support for BasicAuth as an "admin back door" when another Auth provider is used.

PR #1516, opened 2021-05-06
Optional appending tags and updating attributes when correlating alerts

satterly:


satterly:
This PR is blocked waiting on more information.

satterly:
This doesn't solve the original issue and actually introduces breaking behaviour that's difficult to justify. Closing.

Issue #1453, opened 2021-01-22
[SAML] Possibly assign attribute to user.login


gaetanars:
**Is your feature request related to a problem? Please describe.**
When we use SAML auth, the user is created at the first login, we can map attributes to Name and Eamil but we can do this with login value.
When we want to create Api Key to a login its hard to match beetween the user and the actual Login like `AAdzZWNyZXQxGAbtdqeRPPQHdwBm5RaaKmqoSAT4h9IU4H4GudTHYONZDksX2hBe2skMzEjSGCHvUlVKSkiUgUqhJJQoiZ29OCYvzxu7bz2y6hQhM5mIXcARVDR7nJQy0keMEUXTVNAOMfam5odt`

**Describe the solution you'd like**
Add a configuration item like :
`SAML2_LOGIN_ATTRIBUTE: 'identifier'`

satterly:
Closing due to inactivity. If this is still relevant, please reopen with details.

PR #1542, opened 2021-06-07
Add Login Mapping for SAML auth

gaetanars:
This is a proposal to add a SAML attribute for login field in Alerta instead of subject.text.

#1453


satterly:
Is this backward compatible? ie. if someone upgrades Alerta will their login IDs change format?

I would suggest that `SAML2_LOGIN_ATTRIBUTE` be set to `None` by default and if it is unset then `login` will still be `subject.text` otherwise it will use the new setting.

satterly:
Closing due to inactivity. If you believe this issue has been closed prematurely please provide more information to help progress the issue. If you do not know what is expected you can ask on [Slack](https://slack.alerta.dev).

PR #1597, opened 2021-10-26
Do not require scope for metrics view

dakotacody:
When `AUTH_REQUIRED = True`  and AUTH_PROVIDER = 'saml2' this path is not readable without logging in or an API key. 

However, when using telegraf to collect these metrics using the prometheus input plugin, the only types of auth it supports are a bearer token or basic auth.

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/prometheus 

Since /management/healthcheck is similarly exposed with no scope permissions , it's hard to argue /management/metrics should be secured.

### DIFF
--- a/alerta/management/views.py
+++ b/alerta/management/views.py
@@ -224,7 +224,6 @@ def status():
 
 @mgmt.route('/management/metrics', methods=['OPTIONS', 'GET'])
 @cross_origin()
-@permission(Scope.read_management)
 def prometheus_metrics():
 
     now = int(time.time() * 1000)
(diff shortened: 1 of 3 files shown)


satterly:
It makes no difference if the auth token is passed as a header, in the post body or as a URL parameter of you're using https.

satterly:
@dakotacody you have included two unrelated changes in this PR. If you are still interested in either please resubmit as separate PRs. Thanks.

dakotacody:
@satterly   
Seems I caught my mistake in the other PR I made for the grafana settings and reverted that change before making that PR. 
https://github.com/alerta/alerta/pull/1606/files

PR #1606, opened 2021-11-12
Allow grafana webhook to set tags to tags or attributes 

dakotacody:
Many of our plugins rely on tags being set consistently regardless of which system is sending the alert.  

I have added 2 new settings that can be optionally overridden in the default alertad.conf file

`GRAFANA_TAGS_AS_ATTRIBUTES = True`
`GRAFANA_TAGS_AS_TAGS = False`

This will allow for users that are currently using the grafana webhook as it is now to continue using it as is without any configuration changes.   If an enduser wants to have the grafana tags to be set as tags it is now possible to override this and get grafana tags to be tags, attributes, or both if necessary to facilitate migration if someone desires.   

These are set to the above defaults in settings.py 

I considered using a setting like GRAFANA_TAGS = 'attributes' where the config could be set to override to 'tags' and the code would operate one way or the other based off of whether it was set to 'attributes' or 'tags'.   If that is considered better, it could be done that way instead, I just wanted a way to assign to tags on the alert.


satterly:
Thanks! 👍

PR #1637, opened 2021-12-01
add more config environment for docker

oijkn:
USER_DEFAULT_SCOPES
GUEST_DEFAULT_SCOPES
OIDC_GROUP_CLAIM
OIDC_ROLE_CLAIM
OIDC_CUSTOM_CLAIM


satterly:
Thanks for the PR but the config variables that are also supported as environment variables were selected to ease the majority of deployments use them. These variables are for corner-cases that don't follow the "happy path" so I would prefer not to encourage their use.

Issue #1615, opened 2021-11-14
X-Forwarded-User http header


satterly:
Another possibile solution is honoring the X-Forwarded-User http header.  I am running the web UI behind a reverse nginx proxy, which can handle the authorization, and provide the userid via that http header.

_Originally posted by @blysik in https://github.com/alerta/alerta/issues/116#issuecomment-139331798_

satterly:
Done.

PR #1657, opened 2021-12-11
Proxy auth config

### LINKED ISSUE #1578: remote_user authentcation via proxy
I am not sure if its avaliable but with a proxy server server it would be great if I can pass remote_user for authentication.

satterly:
Fixes #1578

### DIFF
--- a/alerta/auth/decorators.py
+++ b/alerta/auth/decorators.py
@@ -120,6 +120,21 @@ def permission(scope=None):
                 else:
                     return f(*args, **kwargs)
 
+            # Proxy auth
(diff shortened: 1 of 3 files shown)


satterly:
Also fixes #1615

satterly:
See #1702 for possible future enhancements.

flinthamm:
Please could someone give an example of how to use proxy-auth, as with no documentation I'm finding this impossible. I'm looking for a way to embed this in an iframe and use token based authentication when the site is viewed locally in an iframe while still be prompted for authentication/login everywhere else. I would be very grateful for any assistance - Thank you.
With customers enabled, I've also created another user but can't login as it seems to need either a customer or * association with all customer records?

Issue #1475, opened 2021-03-05
Implementing delays and generic filter functionality


middelthun:
Hi, Nick!

This will be a quite long post, but please bear with me. We are planning a fairly large PR, so I'd like to give you an early heads up before things are written in stone.

Our company has a need for a plugin that can delay alerts meeting certain (variable) criteria. This has several use cases, but as an example we monitor several services that are occasionally unstable, due to dependencies on infrastructure outside of our own. We want to receive the alerts as they occur, but we don't want them to be active and open until they have existed for a certain amount of time. If something flapped briefly, we don't really care. This could of course be implemented on the monitoring end of things (we mainly use Telegraf and Kapacitor), but given the high amount of stock Telegraf plugins we use, the amount of custom code escalates very quickly if we go down that road. Hence we would much prefer to have this functionality within Alerta.

Pressure to implement this has grown over time, and it has been the most requested feature within our organization for a very long time - which is why we are now dedicating programming resources to fix it.

The basic idea is to implement delays in an almost identical manner as blackouts. CRUD via API for managing delay rules, with an added timeout value per rule that defines how long the alert should be delayed if it matches. When an event reaches Alerta, the plugin will check a table for matching rules, and shelve the alert (for a "timeout" time, as defined by rules) if it matches the criteria. If the alert matches multiple rules, apply avg/min/max (configurable) to the timeout.

However, as we went through the design phase I realised that this should probably be generalised. Duplicating code with just minor differences between them (all those blackout*.py vs delay*.py files) doesn't seem like a very attractive idea.

As a result, we've changed our strategy a bit, and intend to implement this as a generic filter feature.

Each filter looks roughly like the current blackouts, and will indeed be designed so blackouts in time can use the exact same system.

Let me backtrack, and describe it in more detail.

Filters will have a database table, almost identical to the current blackouts table. There can be an arbitrary amount of different filter types (a text column in the table), e.g. "blackout" or "delay". Each filter type can have a set of filter specific data that will be stored in a custom type array (key + value), which can be used by plugins. For instance, start_time, end_time and duration are important for a blackout, but they don't make as much sense for a delay. A delay needs a timeout, to know how long a matching alert should be auto-shelved.

[...]

PR #1660, opened 2021-12-15
feat: Filters

sixcare:
Co-authored-by: Are Schjetne <sixcare.as@gmail.com>

Co-authored-by: Poul H. Sørensen <poul@basefarm.no>

Co-authored-by: Øystein Middelthun <oystein@middelthun.no>


sixcare:
https://github.com/alerta/alerta/issues/1475

sixcare:
missing `filters` https://github.com/sixcare/alerta/blob/feature/filters/alerta/database/backends/postgres/base.py#L146

satterly:
Closing this PR -- the feature concept is sound and addresses a real need (#1475), but the implementation has accumulated too many issues over 4 years to be mergeable. Thanks to the Basefarm team for the significant effort here.

**Key issues:**

1. **Combinatorial matching query** -- `get_matching_filters_by_type` hardcodes every combination of NULL/value for 6 fields (64 OR clauses in both MongoDB and Postgres). This is unmaintainable and performs poorly. The same logic can be expressed as:
   ```sql
   WHERE (resource IS NULL OR resource = %(resource)s)
     AND (service IS NULL OR service <@ %(service)s)
     AND (event IS NULL OR event = %(event)s)
     AND ("group" IS NULL OR "group" = %(group)s)
     AND (tags = '{}' OR tags <@ %(tags)s)
     AND (origin IS NULL OR origin = %(origin)s)
   ```
   That's 6 lines instead of 64, and trivially extensible.

2. **Broken method signature** -- `@staticmethod def find_matching_filters(self, type)` would crash at runtime.

3. **SQL syntax error** in Postgres query builder -- unclosed parenthesis in ID filter.

4. **Wrong column name** -- `get_filter_types` references `UNNEST(types)` but schema defines `type` (singular).

5. **4 years stale** -- conflicts with recent query parser parameterization, dependency updates, and `pkg_resources` removal.

If someone wants to pick this up, the model/view/scope structure here is a good reference for the pattern, but the matching query and backend implementations should be rewritten.

PR #1713, opened 2022-06-27
fix customer group mapping in ldap

Schueni1:
Without this fix groups  from ldap can only be mapped to customers, if and only if there is no domain (email) configured for the user.
With this fix both domain and groups from ldap are considered as customers.


satterly:
Thanks! 👍  🍭

PR #1716, opened 2022-07-01
Add bson dependency, used by utils.format

jjdmol:
This adds bson==0.5.10, which is simply the version pip just gave me.


satterly:
Thanks but it's not the right package. It needs to be the bson package supplied by pymongo.

PR #1752, opened 2022-09-01
Use group parameter of top10 API endpoints

landreville:
The top10 count and top10 flapping API endpoint documentation specifies that a group-by parameter can be provided to group by that value. https://docs.alerta.io/api/reference.html?highlight=top10

It appears that this value is ignored. The `get_topn_flapping` and `get_topn_count` do have a `group` argument, but it is never provided on any invocations.

This pull request gets the first column in the query grouping and uses that for the topn CTE queries. Removing the group argument to get_topn_flapping as unnecessary. I'm not sure if that was the original intention, but the query group is set by the group-by request argument.

This also uses the group column in the CTE selection and in the function output to ensure the JSON output matches the documentation.

### DIFF
--- a/alerta/database/backends/postgres/base.py
+++ b/alerta/database/backends/postgres/base.py
@@ -546,10 +546,14 @@ class Backend(Database):
         """
         return {s.status: s.count for s in self._fetchall(select, query.vars)}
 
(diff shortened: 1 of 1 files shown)


satterly:
thanks!

PR #1799, opened 2023-02-02
Add db schema

Pierre-Louis-Boutruche:
This is an addition to be able to use a specific schema with Postgres. We have modified the `base.py` files to be functional with a specific schema, we have also added the `DATABASE_SCHEMA` environment variable to be able to configure the schema. By default the variable is `none` so the program will take the public schema.


alerta/database/backends/postgres/base.py
```diff
@@ -80,8 +80,10 @@ def create_engine(self, app, uri, dbname=None, raise_on_error=True):
...
         register_adapter(dict, Json)
         register_adapter(datetime, self._adapt_datetime)
+        name = 'history'
+        if schema is not None : name = schema + "." + name
         register_composite(
-            'history',
+            name,
             conn,
             globally=True
         )
```
satterly, line 89: Replace this with ...
```
register_composite(
            schema + '.history' if schema else 'history',
            conn,
            globally=True
        )
```

satterly:
Please address the above comments and add some tests to ensure setting a schema works as expected. Thanks.

satterly:
Btw, you will need to rebase with master so that the tests run.

PR #1814, opened 2023-03-04
Add CLIPBOARD_TEMPLATE setting

emillen:
Only merge this if you decide to merge the PR in the frontend: https://github.com/alerta/alerta-webui/pull/555

**Description**
to allow for configuring this web console setting in the backend: https://github.com/alerta/alerta-webui/pull/555

**Changes**

- Add `CLIPBOARD_TEMPLATE` setting, which can be used to set a template used for formatting the AlertDetail copy-button

---

**Example:**

Provided following alert-json
```json
{
  "text": "this is a major alert",
  "environment": "prod"
}
```
and this template
```python
CLIPBOARD_TEMPLATE = """
[{{environment | upper}}] Alert!
{{text}}
"""
```
would put the following text in your clipboard
```
[PROD] Alert!
this is a major alert
```

### DIFF
--- a/alerta/settings.py
+++ b/alerta/settings.py
@@ -221,7 +221,7 @@ COLUMNS = [
 ]
 SORT_LIST_BY = ['severity', 'lastReceiveTime']  # eg. newest='lastReceiveTime' or oldest='-createTime' (Note: minus means reverse)
 DEFAULT_FILTER = {'status': ['open', 'ack']}
-CLIPBOARD_TEMPLATE = ""
+CLIPBOARD_TEMPLATE = ''
 
 
 # Alert Status Indicators
--- a/alerta/settings.py
+++ b/alerta/settings.py
@@ -221,6 +221,8 @@ COLUMNS = [
 ]
 SORT_LIST_BY = ['severity', 'lastReceiveTime']  # eg. newest='lastReceiveTime' or oldest='-createTime' (Note: minus means reverse)
 DEFAULT_FILTER = {'status': ['open', 'ack']}
+CLIPBOARD_TEMPLATE = ""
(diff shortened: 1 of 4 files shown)


kennhardy (approved):
Looks good. Approving together with https://github.com/alerta/alerta-webui/pull/555

kennhardy (approved):
LGTM

satterly (changes requested):
Please add a test to ensure template config set in the configuration file appears correctly in the API endpoint used by the web UI.

Edit: I can't find a useful example of such a test that you could use as a template so I will write one and push to this branch.

emillen:
Are you working on it? 

Would you be fine with something like this? 

```python
class ApiConfigResponseTestCase(unittest.TestCase):

    def setUp(self):
        test_config = {
            'CLIPBOARD_TEMPLATE': 'this is our template'
        }
        self.app = create_app(test_config)
        self.client = self.app.test_client()

    def test_response_(self):

        response = self.client.get('/config')
        data = json.loads(response.data.decode('utf-8'))
        self.assertEqual(data['clipboard_template'], 'this is our template')

```

satterly:
Yes, that's it. We need to test that config set in the configuration file is correctly returned by the `/config` endpoint.

But the config needs to actually be an example template not just "this is our template". Can you create a test template that uses all the possible template special characters? It would also be a chance to show what it can be used for. People look at the tests when documentation is lacking.

emillen:
I went ahead and added it to the existing `test_config.py`

I hope that will suffice!

satterly (approved):

satterly:
:+1:

PR #1825, opened 2023-03-08
Bulk operations did not work properly

vmgm:
**Description**
Bulk endpoints didn't work because query builder variable was not pointing to `AlertsQueryBuilder`.

Flask login property is needed to create alert history. BG tasks are run in a celery worker so no flask login property is available. 

**Changes**
- fix query builder to use the right one.
- provide login from main process request to celery background task.


satterly:
Thanks! 👍

PR #1826, opened 2023-03-09
Support custom backends defined as entry point

vmgm:
**Description**
A custom backend may be needed to support a different database or to extend or modify current backends functionality. With this modification, a custom backend can be defined using an entry point to provide its module location.
For example, a custom backend based on current Postgres backend, may change the way of correlating alerts. 

**Changes**
- Load backend function to support a backend defined as a package entry-point


alerta/database/base.py
```diff
@@ -29,8 +30,15 @@ def get_backend(app):
 
 
 def load_backend(backend):
+    for ep in iter_entry_points('alerta.backends'):
```
satterly, line 33: Should this be `iter_entry_points('alerta.database.backends')` to be consistent with default db backend location?

satterly: Can you add a minimal example db backend and maybe add it to `examples/backends`?

satterly:
This is very cool. Thanks a lot. 👍

PR #1885, opened 2023-10-09
Address breaking change in Python 3.11 for Enums with `str` mixins

alvanson:
Addresses breaking change in Python 3.11 for `Enum`s with `str` mixins. See https://blog.pecar.me/python-enum for more information on the issue.

Retains compatibility with Python 3.8 by using the `StrEnum` PyPi package rather than the built-in `StrEnum` introduced in 3.11.

This came to my attention when switching from the MongoDB to the PostgreSQL backend under Python 3.11. The JOINs in `get_alerts()` now use the member's name (e.g. `Severity.Minor`) rather than its value (e.g. `minor`) resulting in no alerts showing in the Web UI. The requested changes fix this and likely other issues.

### DIFF
--- a/alerta/models/blackout.py
+++ b/alerta/models/blackout.py
@@ -1,9 +1,9 @@
 from datetime import datetime, timedelta
-from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple, Union
 from uuid import uuid4
 
 from flask import current_app
+from strenum import StrEnum
 
 from alerta.app import db
 from alerta.database.base import Query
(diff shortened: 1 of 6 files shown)


andywow:
Got the same behavior with python3.11 and Postgres

satterly:
Thanks! 👍

mirceaulinic:
👋🏼 thanks @alvanson for this fix. I've been watching this for a while, @satterly is there any ETA for a new release that'll include these changes. Cheers!

Issue #1773, opened 2022-09-26
post_receive | take_action | update attributes behaviour


utsukprani:
Hi,
I have added few custom plugins in my alerta.
1. In one plugin, in the post receive, I am sending the alert to a Kafka queue so that it can be processed by an external system.
2. Then I have this another plugin where I have implemented the take_action. The take_actiion is invoked by the external system after it processes the alert that was went to Kafka. This system sends some attributes which I update on alerta.
3. The external system is also directly calling alerta to create an alert where it populates these attributes.
4. Then I have this another plugin, where in post receive, which is processing the incoming alerts based on the attributes that were sent by the external system (in Point 2) and also in the alert it creates directly in alert via the POST /alert API (in Point 3).

What I am seeing is a bit inconsistent behaviour where sometimes I am loosing the attributes sent by external system.

I have some doubts. Hoping to get some help and clarity 

Question 1 - architecturally is it OK to update the alert attributes in post_receive or is it just meant to notify external system.

Question 2 - as the same alert can come again, which was sent to Kafka, can the use of take_actiin and poat_rexeive cause race conditions or inconsistent behaviour ?

Would appreciate any help. Thanks !

hyberdk:
@utsukprani, Im not entirely sure, but it sounds like you are hitting the same problem that I had.. there was a bug that is the "last" plugin was returning "null" (eg. alert not modified), then it would not update commit any of the previous plusgins changes to alert.. You can checkout #1798 

I hope that solves your problem. Else let me know, as I am also planning to send data of to Kafka in the near future.. ;-)

satterly:
Closing due to inactivity. If this is still an issue, please reopen with details.

PR #1890, opened 2023-12-12
fixes issues if last plugin returns none or raises error

hyberdk:
**Description**
In a scenario where the last plugin configured returns "None" or raises an error the current flow does not update the alert with the information from earlier plugins.

This is only an issue if the very last plugin returns none or raises. Its similar to #1798 and should also fix the issue #1773

**Changes**
basically just ensure that the loop signal that an update are to occur instead of having the last plugin do it.

**Screenshots**
If it's a UI change add screenshots to demonstrate changes.

**Checklist**

**Collaboration**


satterly (approved):
Thanks!

PR #1892, opened 2023-12-19
make action plugins "post_process" instead of "pre_process"

hyberdk:
**Description**
Make actions run after the alert has done its status update to the database. Essentially make the action plugins a "post_process" instead of a "pre_process" plugin

Fixes # (issue)

**Changes**
Include a brief summary of changes...

- I have swapped the order, so that whenever an action is executed via the API the alert is updated in the database including setting the updated status (if ack/unack/close/open/expire action) before the alert is sent to the plugins. Today the action is received sent to the plugins (with the previous status) and then once the plugins are done running it updates the status in the alerts (and the database)

**Checklist**

**Collaboration**

### DIFF
--- a/alerta/views/alerts.py
+++ b/alerta/views/alerts.py
@@ -156,8 +156,8 @@ def action_alert(alert_id):
         raise ApiError('not found', 404)
 
     try:
-        alert, action, text, timeout = process_action(alert, action, text, timeout)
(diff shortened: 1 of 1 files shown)


hyberdk:
will add a "post_take_action" method to the PluginBase instead, that should not be a breaking change.

PR #1899, opened 2023-12-21
feat: implement "post_action" for post processing actions.

hyberdk:
**Description**
This feature add the option to implement an "post_take_action" plugin. The problem is that the current action plugins runs before the alert as its status_update and the alert changes status.

If you want an plugin that can do something _after_ that the action is taken, then today its not really possible or very tricky to do. 

This is similar to the pre_recieve / post_recieve setup.

This is also backwards compatible with existing plugins.

**Changes**
Implements the "post_take_action" method in the PluginBase, and then run it after the alert is updated

**Checklist**
- [X] Added new tests related to change

**Collaboration**


tests/test_plugins.py
```diff
@@ -557,6 +568,12 @@ def take_action(self, alert, action, text, **kwargs):
 
         return alert, action, text
 
+    def post_take_action(self, alert, action, text, **kwargs):
+
+        if action == 'test-post-action':
+            alert.tags.append('test-post-action')
```
satterly, line 574: It would be better to have something a bit more realistic as the `post_action` hook. What is a use-case that you currently have for post actions? Does it fit with this example of "createTicket", "updateTicket", "resolveTicket"?

satterly (changes requested):
LGTM. Just a few minor suggested improvements. Thanks!

tests/test_plugins.py
hyberdk: My use case is actually that it triggers some external event. In our case its pushing a new event to a kafka topic that the "action" was completed, so I did not really want to put that into the test-case. I just wanted to make sure that the function was triggered.

alerta/plugins/__init__.py
```diff
@@ -35,7 +35,16 @@ def status_change(self, alert: 'Alert', status: str, text: str, **kwargs) -> Any
...
 
     def take_action(self, alert: 'Alert', action: str, text: str, **kwargs) -> Any:
-        """Trigger integrations based on external actions. (optional)"""
+        """
+        Trigger integrations based on external actions. (optional)
+        Pre-trigger, eg. this triggers before the status are updated.
+        """
+        raise NotImplementedError
+
+    def post_take_action(self, alert: 'Alert', action: str, text: str, **kwargs) -> Any:
```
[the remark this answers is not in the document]
hyberdk: sure, no problem. I will update the PR

satterly:
Thanks! 👍

Issue #1884, opened 2023-10-09
unable to query alerts api for more than 2 tags. / params


streaming-pete:
**Issue Summary**
When trying to filter results via the API (or UI) using tags, we are unable to use more than 2 tags.

**Environment**
- OS: Linux
- API version: 9.0.1
- Deployment: Local / K8s
- For self-hosted, WSGI environment: nginx / pycharm

- Database: Postgres & Mongo

- Server config:
  Auth enabled? Yes for Prod, No for Dev
  Customer views? No

- web UI version: n/a
- CLI version: n/a

**To Reproduce**
create 3 alerts with unique tags. eg
 alerta  send -r test4 -e test4 -S test4 -t test4 -T dc=tag1 -E Production -s major
 alerta  send -r test3 -e test3 -S test3 -t test3 -T dc=tag2 -E Production -s major
 alerta  send -r test2 -e test2 -S test2 -t test2 -T dc=tag3 -E Production -s major

query api with following params

/alerts?q=tags%3A"dc%3Dtag1"+tags%3A"dc%3Dtag2"+tags%3A"dc%3Dtag3"

response will only be for tag1 + tag2.

code generates: **"('dc=tag1'=ANY("tags") OR 'dc=tag2'=ANY("tags"))"**

**Expected behavior**
all tags in search returned

**"('dc=tag1'=ANY("tags") OR 'dc=tag2'=ANY("tags") OR OR 'dc=tag3'=ANY("tags"))"**

**Screenshots**

**Additional context**
This was a customer request from our ops dept using alerta.

I have done some debugging and testing against mongo and postgres, in debug mode I can trace the params correctly getting to the QueryParser class in queryparser.py and then hitting the query_expr infixNotation() that seems to only handle a left / right or clause.      (Optional(OR).setParseAction(lambda: 'OR'), 2, opAssoc.LEFT, SearchOr), this in turn returns an sql query of: 

**"('dc=tag1'=ANY("tags") OR 'dc=tag2'=ANY("tags"))"**

so it looks like the parser for handling multiple query params is locked at 2 even though 3 params were passed into the function.

Appreciate any feedback or guidance on what we could do here to open this up to more parameters for searching.

Thanks
Pete

streaming-pete:
Interestingly, I can get up to 4 working by wrapping 2 sets of OR's in brackets via the UI. (I have moved to the tag called channel here which are in my prod system i have lots alerts for)

[...]

PR #2020, opened 2025-02-17
Query more than two fields with Postgres backend

Titof974:
**Description**
Postgres backend doesn't support more than two parameters in the query.

For example if I request: 
`origin:"myservice" AND severity:"MAJOR" AND _.key:"value2"` Alerta will generate this request `"origin" ~* '\myservice\y' AND"severity" ~* '\yMAJOR\y'` and skip the `_.key:"value2"` part.

If I change the order of the request to ` _.key:"value2" AND origin:"myservice" AND severity:"MAJOR"`, it will generate  `"attributes"::jsonb ->>'key' ~* '\yvalue2\y' AND "origin" ~* '\myservice\y'` skipping the `severity:"MAJOR"` part.

Same behavior with the OR condition where `origin:"myservice" AND severity:"MAJOR" OR !_.key:"valueAAA"` is transformed into `"origin" ~* '\myservice\y' OR "severity" ~* '\yMAJOR\y'`.

Fixes # (issue)
https://github.com/alerta/alerta/issues/1884

**Changes**
Handle multiple AND/OR operators in Postgres query parser.

**Checklist**
- [ ] Pull request is limited to a single purpose
- [ ] Code style/formatting is consistent
- [ ] All existing tests are passing
- [ ] No unnecessary whitespace changes

### DIFF
--- a/alerta/database/backends/postgres/queryparser.py
+++ b/alerta/database/backends/postgres/queryparser.py
@@ -16,9 +16,7 @@ class BinaryOperation:
     """takes two or more operands, e.g. and, or"""
 
     def __init__(self, tokens):
-        self.op = tokens[0][1]
-        self.lhs = tokens[0][0]
-        self.rhs = tokens[0][2]
+        self.tokens = tokens
 
 
 class SearchModifier(UnaryOperation):
@@ -30,15 +28,13 @@ class SearchModifier(UnaryOperation):
 class SearchAnd(BinaryOperation):
 
     def __repr__(self):
-        return f'({self.lhs} AND {self.rhs})'
+        return " AND ".join(map(repr, self.tokens[0]))
 
 
 class SearchOr(BinaryOperation):
 
     def __repr__(self):
-        if getattr(self.rhs, 'op', None) == 'NOT':
-            return f'({self.lhs} AND {self.rhs})'
-        return f'({self.lhs} OR {self.rhs})'
+        return " OR ".join(map(repr, self.tokens[0]))
 
 
 class SearchNot(UnaryOperation):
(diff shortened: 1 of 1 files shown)


Yvanohe:
Hello @Titof974 ,
I may have found an issue with your fix. It seems to flatten query :
A AND (B OR C) => A AND B OR C

Example request :
` q=_.Country:"France" AND _.alert_category:("Battery issue" OR "Sensor issue" OR "Test") `

will generate with your **fix** :  
`
"attributes"::jsonb ->>'Country' ~* '\yFrance\y' AND "attributes"::jsonb ->>'alert_category' ~* '\yBattery issue\y' OR "attributes"::jsonb ->>'alert_category' ~* '\ySensor issue\y' OR "attributes"::jsonb ->>'alert_category' ~* '\yTest\y'`
 -> All parameters are kept but the AND and OR structure not respected

will generate with the **current** Postgres backend : 
`("attributes"::jsonb ->>'Country' ~* '\yFrance\y' AND ("attributes"::jsonb ->>'alert_category' ~* '\yBattery issue\y' OR "attributes"::jsonb ->>'alert_category' ~* '\ySensor issue\y'))` 
-> AND and OR structure respected but miss one parameter

I will look to a solution in your code

satterly:
Closing this PR -- the problem it tries to solve (queries with more than two AND/OR terms) already works correctly via pyparsing's `infixNotation`, which builds nested binary trees. For example, `a AND b AND c` parses as `((a AND b) AND c)`, which produces valid SQL.

The changes here fundamentally alter the query parser in ways that break existing behavior:

1. **NOT handling is broken** -- `SearchOr.__repr__` currently checks `getattr(self.rhs, 'op', None) == 'NOT'` to convert `a OR NOT b` into `a AND NOT b`. With the flat token list approach, `self.rhs` no longer exists, so this logic is lost.

2. **Suppressing AND/OR keywords** from the token stream means the operator type information is discarded -- `SearchAnd` and `SearchOr` both just join all remaining tokens, but can't distinguish between operands and operators if the grammar ever produces mixed tokens.

3. This also conflicts with recent changes to the query parser (parameterized queries in fdd52cd) which rely on the binary tree structure with `self.lhs`/`self.rhs` and `to_sql()` methods.

If you're hitting a specific query that doesn't work with more than two terms, please open an issue with the exact query string and we can investigate.

Thanks for the contribution.

PR #2075, opened 2026-09-22
chore: adapt fork to this organisation's infrastructure and release process

ospo-ionik:
Repoints everything the fork inherited from upstream at our own infrastructure, so CI runs, releases are publishable, and no artefact claims to be upstream's.
- **CI** — workflows now trigger on `main` (PRs previously ran no tests at all); removed Slack notifications and the fly.io deploy of upstream's demo instance.
- **Packaging** — distribution renamed to `alerta-ng-server`; the importable package stays `alerta`, so plugins and deployments are unaffected. Version unified on `alerta/version.py`.
- **Releases** — PyPI publishing moved to Trusted Publishing (no long-lived token), with a guard that the tag matches the version. Removed `make upload`, which bypassed CI.
- **Metadata** — container labels, image references, project URLs and README badges now point here. Upstream attribution in `LICENSE` and `NOTICE` is untouched.
- **Version** — set to `10.0.0` so we never reuse a number from upstream's line. Not a compatibility break.


ospo-ionik:
Please, ignore and delete

